PR #6, opened 2025-09-14
Fixed dev container rviz build error + mounting

Rain1618:
Problem: Dev container wasn't building 
Changes made:
- Removed ros-humble-rviz (does not exist) and replaced with the correct ros-humble-rviz2 package.
- Added dependencies from the  Jetson container. Some of them may be unnecessary/we might want other packages... 
- Added proper bind mounting of the AUV-2026 repo to simplify Git usage inside the container.
- Updated workspace folder handling so builds and sources happen in the correct place.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,17 +6,20 @@
 			"WORKSPACE": "${containerWorkspaceFolder}"
 		}
 	},
-    // This will launch the container as a non-root user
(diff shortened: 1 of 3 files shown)


Alb33rt (approved):
Next time the branch must have the prefix `dev/*` but otherwise the changes look reasonable. Definitely did not realize that there were issues with the code rn. Approved.

PR #9, opened 2025-09-19
movella IMU into main

### LINKED ISSUE #3: Integrate Movella IMU
## 🎯 Objective
Integrate the Movella IMU sensor with our ROS2 navigation stack to provide reliable orientation, acceleration, and angular velocity data for state estimation and motion control.

## ✅ Acceptance Criteria
- [x] Movella IMU ROS2 package compiles successfully with colcon build
- [x] Able to rostopic echo /imu/data and see valid IMU data (orientation, linear acceleration, angular velocity)
- [ ] IMU data is properly transformed to robot's base_link frame (auv)
- [x] Covariance matrix is properly populated in IMU messages (take the same covariances as in AUV-2025, those values are calculated using movella's documentation)
- [x] Udev rules created that assign a consistent symlink (e.g., /dev/movella_imu) regardless of which USB port is used

## 🛠️ Implementation
**Approach:** 
1. Create udev rules for the IMU and then reload the udev rules (if you do `ls /dev/movella_imu`, there should be a device)
- `sudo cp [path to your rule]  /etc/udev/rules.d/ `
- `sudo udevadm control --reload-rules && sudo service udev restart && sudo udevadm trigger`
- see [this](https://docs.clearpathrobotics.com/docs/ros1noetic/ros/ros/tutorials/ros101/intermediate/udev_rules/) for more info about udev rules
3. Take a look at [the official Xsens driver](https://github.com/xsenssupport/Xsens_MTi_ROS_Driver_and_Ntrip_Client/tree/ros2) 
4. Make sure all dependencies (such as nvea-msgs) are in our Dockerfile. If not, add and rebuild (have fun, bring snacks, it might take time)
5. Follow all the steps in the README.md and make sure you can echo data 
## 🔗 Links
**Blocks:** Movella republisher REP 103/105, State estimation
**Blocked by:** https://github.com/mcgill-robotics/AUV-2026/issues/1

WilliamZhang2205:
# Closes #3  
- Movella IMU no in our ros2 :(

# Fix
- Converted movella_republish.cpp which removes gravity acceleration
- Added CMakeLists.txt for packages
- Changed ament_python to ament_cmake to compile cpp files
- Added Launch file to create Movella Node and launch Xsens Node
- Made a gitmodule fork to allow change of movella params

# Tests Ran
- Launched imu_movella_launch.py and see if nodes are there with ros2 node list
- Echoed topics of sensor/imu/data and moved imu around 
- Checked covariances in the topic message being published


Rain1618:
Proposal: we make a clone of the driver using the mcgill robtics account. we make any changes needed there and add that repo as a submodule

Rain1618 (dismissed):
Code looks good, if everything has been tested on the nano and works, merge.

ros2_ws/src/sensors/launch/imu_movella_launch.py
```diff
@@ -0,0 +1,25 @@
...
+            PythonLaunchDescriptionSource(['/root/ros2_ws/install/xsens_mti_ros2_driver/share/xsens_mti_ros2_driver/launch', '/xsens_mti_node.launch.py']))
+
+    return LaunchDescription([
+       included_launch,
+        Node(
+            package='sensors',
+            executable='imu',
+            name='movella_imu',
+            parameters=[
+                {"frame_override":"imu"}
```
Alb33rt, line 21: Not sure if frame overriding here is a great idea, we need to make sure this setting is consistent between here and the sim. Just curious as to why.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
+#include <rclcpp/rclcpp.hpp>
+#include <sensor_msgs/msg/imu.hpp> 
+#include <geometry_msgs/msg/vector3_stamped.hpp>
+#include <message_filters/subscriber.h>
+#include <message_filters/sync_policies/approximate_time.h>
+#include <message_filters/synchronizer.h>
+#include <message_filters/time_synchronizer.h>
+
+class MovellaRepublisher : public rclcpp::Node
```
Alb33rt, line 9: In C++, don't just add the structure and details of the node in the `.cpp` file. Make a `.hpp` file and include the interface of the class over there (this includes fields). The `.cpp` file should only be concerned with the details of functions.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+			// Publish new message
+			pub_imu->publish(processed_imu_msg);
+		}
+
+		rclcpp::Publisher<sensor_msgs::msg::Imu>::SharedPtr pub_imu;
+		message_filters::Subscriber<sensor_msgs::msg::Imu> sub_imu;
+		message_filters::Subscriber<geometry_msgs::msg::Vector3Stamped> sub_free_acc_imu;
+		std::shared_ptr<message_filters::Synchronizer<message_filters::sync_policies::ApproximateTime< \
+			sensor_msgs::msg::Imu, geometry_msgs::msg::Vector3Stamped>>> sync;
+		std::string frame_override;
```
Alb33rt, line 75: We're missing a constructor class which might render `this->sync` null, causing a memory leak. Online implementations tend to wrap `message_filter` sync type inside a `std->shared_ptr`. Look into **smart pointers** as they are a staple in modern C++. This applies to all the fields as well. Try to make them smart_ptrs if they are not primitive types. These are why memory leaks happen irl.

Example [here](https://docs.ros.org/en/rolling/p/message_filters/doc/Tutorials/Approximate-Synchronizer-Cpp.html#:~:text=private%3A%0A%20%20%20%20rclcpp,SharedPtr%20second_timer%3B)

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+};
+
+
+int main(int argc, char *argv[])
+{
+	rclcpp::init(argc, argv);
+	rclcpp::spin(std::make_shared<MovellaRepublisher>());
```
Alb33rt, line 82: This is an actual example of how to make a `shared_ptr` using `std`.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+			// Prepare new data content 
+			processed_imu_msg.orientation  = imu_msg->orientation;
+			processed_imu_msg.orientation_covariance = imu_msg->orientation_covariance;
+			processed_imu_msg.angular_velocity = imu_msg->angular_velocity;
+			processed_imu_msg.angular_velocity_covariance = imu_msg->angular_velocity_covariance;
+			processed_imu_msg.linear_acceleration = free_acc_msg->vector;
+			processed_imu_msg.linear_acceleration_covariance = imu_msg->linear_acceleration_covariance;
+
+			// Check if the two fused messages are too far apart time wise
+			double dt = fabs((imu_msg->header.stamp.nanosec - free_acc_msg->header.stamp.nanosec) * 1000000);
```
Alb33rt, line 58: I'm not sure if taking the literal nanosec and multiplying to 10^6 is a great idea. Might require more explanation.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+		void callback(const sensor_msgs::msg::Imu::ConstSharedPtr& imu_msg, 
+				const geometry_msgs::msg::Vector3Stamped::ConstSharedPtr& free_acc_msg) const
+		{
+
+			// Takes imu_msg and imu without acceleration message and fuse their data into one message
+			// for state estimation
+			 
+			
+			// Prepare new header
+			sensor_msgs::msg::Imu processed_imu_msg;
```
Alb33rt, line 44: When instantiating beyond C++17, try to use the newer constructors that leverages smart pointers. This would look like `sensor_msgs::msg::Imu::SharedPtr processed_imu_msg = std::make_shared<sensor_msgs::msg::Imu::SharedPtr>();` or something like that. This is to prevent this message type from being passed on to something else, and then when `MovellaRepublisher` gets destructed, the message reference would still point here! This causes memory leak. For more reference check [here.](https://stackoverflow.com/questions/106508/what-is-a-smart-pointer-and-when-should-i-use-one)

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+			processed_imu_msg.orientation_covariance = imu_msg->orientation_covariance;
+			processed_imu_msg.angular_velocity = imu_msg->angular_velocity;
+			processed_imu_msg.angular_velocity_covariance = imu_msg->angular_velocity_covariance;
+			processed_imu_msg.linear_acceleration = free_acc_msg->vector;
+			processed_imu_msg.linear_acceleration_covariance = imu_msg->linear_acceleration_covariance;
+
+			// Check if the two fused messages are too far apart time wise
+			double dt = fabs((imu_msg->header.stamp.nanosec - free_acc_msg->header.stamp.nanosec) * 1000000);
+
+			if (dt > 0.015)
```
Alb33rt, line 60: if possible make the time difference a dynamic parameter that can be changed during launch time instead of compilation time.

ros2_ws/src/sensors/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+add_executable(imu sensors/movella_republish.cpp)
+ament_target_dependencies(imu rclcpp std_msgs geometry_msgs sensor_msgs  message_filters)
+
+install(TARGETS
+	 imu
+	DESTINATION lib/${PROJECT_NAME})
+
+install(DIRECTORY
+	 launch
+	DESTINATION share/${PROJECT_NAME})
```
Alb33rt, line 29: This is actually why we need the header file! (Refer to my other comment about headers) The public exposure for the library should only be the `.hpp` file as the exact implementation would traditionally vary OS-to-OS. This is just a good SWE practice to have the `.hpp` file be in `share/` instead of the actual source file.

ros2_ws/src/sensors/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+project(sensors)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+	  set(CMAKE_CXX_STANDARD 17)
+  endif()
+
+  if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+	    add_compile_options(-Wall -Wextra -Wpedantic)
+    endif()
```
Alb33rt, line 11: indentiation. also -Wextra tag for `clang` adds deprecation warning on dependencies that may trigger us if we use older libraries.

Alb33rt (changes requested):
Great first C++ task. There are a few things to be addressed that I labelled. Sorry if this may seem like trouble, but its for the betterment of our SWE practices.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,91 @@
...
+	{
+		// Create subscribers and publisher
+		pub_imu = this->create_publisher<sensor_msgs::msg::Imu>("sensor/imu/data", 20);
+		sub_imu.subscribe(this, "imu/data"); //uses .subscribe instead of create_subscription because using messageFilters
+		sub_free_acc_imu.subscribe(this, "filter/free_acceleration");
+	
+
+		// Create a synchronizer that synchs the free_acc message and regular imu message 
+		// with an interval of 0.05s allowed time and calls the callback func to process the messages
+		using SyncPolicy = message_filters::sync_policies::ApproximateTime< \
```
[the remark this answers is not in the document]
WilliamZhang2205: We'd have to ask Nadim for deep reasoning. I saw it as the two messages may not arrive exactly on the same time. So youd create a policy where you match the two closest messages in time with the message_filters algorithm. 0.05s is the period of the publisher so i assumed thats how it worked. Ill ask Nadim about it.

ros2_ws/src/sensors/CMakeLists.txt
Alb33rt: We still need to add the header files from with `include/sensors/` to part of the compilation process because it is  in the usable interface.

Alb33rt (changes requested):
We still need to add the header files to part of the compilation process. Everything else looks great! Great job!

Rain1618 (dismissed):

ros2_ws/src/sensors/CMakeLists.txt
Alb33rt: Indentation here is messed up alongside the `DIRECTORY` names. I know it's nit-picky but the convention is to use `launch/` (or append `/` to denote that it is a directory and not a target name) and the target name for `include` should be `include/`. This applies to the `install` function below this line as well if you change that to target-based.

ros2_ws/src/sensors/sensors/movella_republish.cpp
```diff
@@ -0,0 +1,85 @@
+#include <rclcpp/rclcpp.hpp>
+#include <sensors/imu_movella.hpp>
```
Alb33rt: The header file should have the same name as the `.cpp` file.

Alb33rt (changes requested):
`include/` header file inclusion code requires some changes! I was also curious as to why we have package metafiles overlap for both an `ament_python` package and `ament_cmake` package. Code format is off in some places (just a heads up, change not required, I'll fix this in one PR later). Pull main into your branch first.

Everything else looks good. Solid code logic. The `C++` building system can get to the brains real quick.

Alb33rt (approved):
PR looks good. Solid C++ code with constructor and destructor. Has been tested on the Jetson as well. Approved.

Rain1618 (approved):
nice work

PR #16, opened 2025-10-11
Added custom AUV messages in ROS2

JPGC04:


Rain1618 (approved):
We should consider removing some message types if they are not needed in the future

PR #22, opened 2025-10-25
Add colcon build CI/CD and unity simulation tcp package

Alb33rt:
# 🔑 Features
- Added `colcon build` and baseline testing **CI/CD pipeline.**
- Resolved CMake issues with sim tcp endpoint.
- Added working simulation endpoint package.

# 🔧 Fixes
- Additional `ament_cmake` dependencies
- `build.sh` instead of `colcon build`

# ⁉️ Potential Questions
- How well CI pipeline performs on other branches. This can only be verified efficiently by trying it out on other branches.

# Tests Ran
- Github Actions passes on local branch when building
- Dockerfile compilations passes

### DIFF
--- a/ros2_ws/src/auv_msgs/CMakeLists.txt
+++ b/ros2_ws/src/auv_msgs/CMakeLists.txt
@@ -1,0 +1,47 @@
+cmake_minimum_required(VERSION 3.8)
+project(auv_msgs)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+set(MSG_DEP_SET 
+	geometry_msgs 
+	std_msgs
+	actionlib_msgs
+)
+
+set(msg_files	
+	"msg/DeadReckonReport.msg"
+	"msg/PingerBearing.msg"
+	"msg/PingerTimeDifference.msg"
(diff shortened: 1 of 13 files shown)


Docker/jetson/Dockerfile
```diff
@@ -197,14 +197,14 @@ RUN git clone --branch ros2 https://github.com/xsenssupport/Xsens_MTi_ROS_Driver
 
 # Build the ZED Wrapper and IMU driver
```
WilliamZhang2205: In the most recent jetson Dockerfiles, I opted to not build the zed ros2 wrapper and imu drivers since they were "ignored" when you mounted the container to the local repo. This lead to some confusion in development. For example, say you decided to build zed ros2 wrapper but the local repo did not have it. When you got inside the container there would be nothing and youd be left confused as to why your dockerfile Colcon build did not work but it did you just mounted over it. Instead, the zed ros2 wrapper and IMU drivers have been built on the local repo and they just live there. If you think thats a bad idea, please tell.

Alb33rt: Building on the local repository is a bad idea because you're using different version of everything that you need to maintain the consistency of. That sort of overrides the whole point of the Dockerfile.

When you got inside the container and there was nothing, it was probably you forgot to git clone recursively with the submodules. Lmk if this wasn't the issue. I commented out the building here because I think we should build manually before our drytest.

Rain1618 (approved):

PR #23, opened 2025-10-26
Add custom ROS2 messages

JPGC04:


Rain1618 (approved):
As per convo w/ JP, some messages will not be used, only merging them due to the bridge and they will be phased out in the future

Issue #26, opened 2025-10-28
Establish clear structure and logic for image enhancement decisions based on experimental algorithms


chantalzhang:
## 🎯 Objective
Clean up and structure experimental image enhancement code to make it ready for integration into the vision pipeline.

## 📖 Description
This task focuses on cleaning up and organizing the existing image enhancement code, which was originally developed through experimentations (see issue #13 and PR #14 for context). The goal is to establish a clearer structure and consistent logic that is ready to be integrated into the vision pipeline. This includes narrowing down from all the experimental algoritms, ensuring the code is maintainable and ready for refactoring into vision.

## ✅ Acceptance Criteria
- [ ] Refactor experimental image enhancement scripts into organized, modular classes for maintainability and reuse
- [ ] Implement caching or lightweight state management where appropriate for performance
- [ ] Remove ineffective or irrelevant algorithms (e.g. white patch method)
- [ ] Clean up unused parameters, redundant logic, and leftover experimental code
- [ ] Ensure the new class structure is ready for integration into the AUV vision pipeline
- [ ] Update README.md quickly so we can easily move this to the wiki when refactoring into vision 

## 🛠️ Implementation
Step-by-step workflow: 
-Review current code: Identify experimental parts, unused parameters, and hardcoded values.
-Create/organize classes: Make separate classes for each enhancement type and keep them simple (some classes do not need __init__, such as HistogramEqualization, UnderwaterColorCorrection)
-`enhancement_comparison.py` should probably be removed in this PR, or definitely the next 
-Document quickly changes and choices made 

## Next 
- Next PR will be to refactor into the vision pipeline, where we will remove README.me and add the updated one after iterating to the wiki and remove it from the codebase.

PR #14, opened 2025-09-25
Image enhancement experimenting

chantalzhang:
Addresses and resolves issue #13 . 

Can choose to .gitignore the input/output pictures, but for now there are only two pictures, and in the image enhancement folder. Once we select which and how we want to implement the image enhancement algorithms, we can refactor this subfolder to integrate it with the rest of the pipeline. 

To run the testing script: 

You can upload any input picture into the input subfolder. 

From AUV-2026: 
`cd image_enhancement`
`python enhancement_comparison.py`

The tool will save a single comparison image in the output/ folder. This image displays the original input alongside the nine enhanced versions generated by different algorithms. For more details on each algorithm, refer to the `README.md`

PR #27, opened 2025-10-29
Integrate image enhancement script 

chantalzhang:
Addresses and resolves issue #26. Reworked the experimental image enhancement #14 into a usable pipeline ready to be refactored into vision. 

It supports multiple enhancement algorithms, organized into four categories:

1. Color Correction: WhiteBalance, RedChannelEnhancement, UnderwaterColorCorrection  
2. Backscatter Reduction: DCPEnhancement  
3. Edge Preservation: GuidedFilter  
4. Contrast Enhancement: CLAHEEnhancement, GammaCorrection, HistogramEqualization 

The current default pipeline applies the following sequence:
1. WhiteBalance (color correction)  
2. DCPEnhancement (backscatter reduction)  
3. GuidedFilter (edge preservation)  
4. CLAHEEnhancement (contrast enhancement)  
  
This will be revised after testing, but we believe this order ensures each step builds on the previous one:
- **WhiteBalance** corrects color casts first for accurate colors.  
- **DCPEnhancement** clears haze and backscatter to reveal detail.  
- **GuidedFilter** reduces noise while keeping edges sharp.  
- **CLAHEEnhancement** improves contrast for a clearer, more balanced image.

### Changes Made since experimental branch 
**Architecture**
- Refactored from experimental scripts to object-oriented classes for maintainability and caching.  
- Added `EnhancementAlgorithm` base abstract class for a consistent API.  
- Implemented composite `ImageEnhancer` class to support algorithm chaining.  

**Cleanup and Optimization**
- Removed unreliable White Patch method (misinterprets water reflections).  
- Removed unused or redundant code (`UnsharpMasking`, unnecessary `__init__` methods).  
- Reorganized algorithms by category with clear headers.  
- Removed outdated README.md, comparison script and requirements.txt used during experimentation.  

### Next Steps
Integrate the enhancement pipeline into the ROS vision system:
- Subscribe to ZED Camera image topics.  
- Convert images using `cv_bridge` to NumPy arrays.  
- Apply the `ImageEnhancer` pipeline and republish enhanced images.  

This approach is faster and cleaner than using PyTorch transforms, reduces dependencies, and keeps enhancement decoupled from detection for easier refactoring. Future optimizations can include GPU acceleration via OpenCV (`cv::cuda`) or a C++ ROS node implementation.

### DIFF
--- a/image_enhancement/image_enhancement.py
+++ b/image_enhancement/image_enhancement.py
@@ -1,167 +1,199 @@
-
 import cv2
 import numpy as np
 from typing import Tuple
 from abc import ABC, abstractmethod
 
+
 class EnhancementAlgorithm(ABC):
-        '''Abstract Base Class containing various image enhancement algorithms for underwater images.'''
-        @abstractmethod
-        def apply_algorithm(self, image: np.array) -> np.ndarray:
-                pass
+    """Abstract Base Class containing various
+    image enhancement algorithms for underwater images."""
 
-        def __call__(self, image: np.ndarray) -> np.ndarray:
-                return self.apply_algorithm(image)
-        
+    @abstractmethod
+    def apply_algorithm(self, image: np.ndarray) -> np.ndarray:
+        pass
+
+    def __call__(self, image: np.ndarray) -> np.ndarray:
+        return self.apply_algorithm(image)
+
+
 class ImageEnhancer:
-        '''Aggregation of enhancement algorithms'''
-        def __init__(self,*algorithms):
-                # Current combination and ordering of algorithms for image enhancement (one of each of the categories )
-                if not algorithms: 
-                        self.algorithms = [
-                                WhiteBalance(),                   
-                                DCPEnhancement(),                 
-                                GuidedFilter(),                    
-                                CLAHEEnhancement()                 
-                        ]
-                else:
-                        self.algorithms = list(algorithms)
+    """Aggregation of enhancement algorithms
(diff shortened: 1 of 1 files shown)


Docker/dev/Dockerfile
Rain1618: Have we tested these algorithms on the jetson as well? I think we have opencv in the jetson dockerfile as well but if not (or if there's version incompatibility issues) we can add it when we write the actual image_enhancement node

Rain1618 (dismissed):
This looks great guys!!!!!

Docker/dev/Dockerfile
avyan-k: No, I don't think we have tested them on the jeston yet. From memory, the algorithms that require `opencv-contrib` (and thus the newer numpy version) are `WhiteBalance` and `GuidedFilter`. So running something like this
```python
import image_enhancement as ie
# the two algos we can test for now, set to default params
enhancer = ie.ImageEnhancer(ie.DCPEnhancement(),ie.CLAHEEnhancement())
# img is some image in BGR/opencv format
enhanced_img = enhancer(img)
```
Should be good for a test. I can do it but I may need some help with figuring out the jetson. As you mentioned we can also just wait for the ROS node branch and test on that instead.

Alb33rt (approved):
Great changes and great software engineering principles. Excellent results on the image enhancement results as well. We're likely going to be ready for object detection after.

PR #30, opened 2025-11-04
noetic base + humble base w ros1_bridge and custom_msgs

JPGC04:


Rain1618 (approved):
Will approve but as per discussion with JP, we need to phase this out and start using the actual jetson docker in ros2 for tests

Alb33rt (approved):
Agreed with Raina on the approach for testing directly in the ROS2 container instead. But will approve due to pooltesting needs rn.

PR #43, opened 2025-12-21
[WIP] Depth and IMU data processing

nadimasmar:
This PR restructures the sensors package and adds processing for incoming IMU and depth data.

1. Added `ImuRepublisher` and `DepthRepublisher` processors and integrated them into a single `sensor_node` executable for easier state sharing (e.g., using the latest IMU orientation when processing depth).

2. Updated CMakeLists.txt to build/install sensor_node and removed obsolete executables/sources from the package.

3. Updated launch to start the Xsens driver only when sim:=false, `sensor_node` runs regardless of sim. 

4. Updated package.xml to include Eigen dependencies and refreshed README.md with detailed IMU/depth processing documentation.

TODO:
- Add a DVL data processor. 
- decide on a sensor fusion procedure or a simple state aggregator
- Test and iterate


JPGC04 (dismissed):
Very nice progress in this pr. All this needs to be tested in real life to make sure data is actually being published in real life.

WilliamZhang2205 (commented):
Minor stuff in comments I left. Great BOOMIN work overall. I agree with what the big boss JP man say, what he said

Alb33rt (dismissed):
Changes verified via real-world testing. Suggest changing the name of the quaternions because they aren't very readable for less mathematically inclined engineers.

ben-jamming (changes requested):

ros2_ws/src/sensors/CMakeLists.txt
```diff
@@ -1,43 +1,72 @@
...
+install(DIRECTORY 
+        params/
+        DESTINATION share/${PROJECT_NAME}/params
+)
 
 install(DIRECTORY
-	include/
-	DESTINATION include/${PROJECT_NAME})
+        include/
+        DESTINATION include/${PROJECT_NAME})
```
kaidalisohaib: For `launch/`, the trailing slash causes the files to be installed into `share/sensors/` instead of `share/sensors/launch/`.

For the `include` part, the headers file are being installed to `include/sensors/sensors` instead of the wanted `include/sensors`.

Something along these lines should fix those points:

```
install(DIRECTORY
        launch
        DESTINATION share/${PROJECT_NAME})

install(DIRECTORY 
        params
        DESTINATION share/${PROJECT_NAME}
)

install(DIRECTORY
        include/
        DESTINATION include)
```

kaidalisohaib (dismissed):

PR #44, opened 2025-12-28
Dev/controls

nadimasmar:
This PR adds controls and state estimation packages. Both packages have been tested in the pool and provide the following functionality:

1. IMU data transformation into the AUV frame. Extensively tested in the pool
2. DVL data transformation into the AUV frame. Modereately tested in the pool.
3. Depth Sensor data transformation into the AUV frame. Extensively tested in the pool
4. Depth Control. Dive, hover, and surface capability that is accurate and precise. Extensively tested in the pool. 
5.  Attitude Control. Yaw accurately and precisely to any possible heading. Corrective Roll and Pitch in case of a disturbance. Roll and Pitch maneuvers (such as barrel rolls) are not supported, only corrections to the neutral state in case of a disturbance.  Extensively tested in pool
6. Planar ( X & Y) control. Translate along the X and Y axes accurately and precisely. Lightly tested in the pool. Only straight maneuvers have been tested, diagonal movements have not been tested.


ros2_ws/src/controls/controls/pid.py
```diff
@@ -0,0 +1,34 @@
...
+                self.integral_error = 0.0
+                self.derivative_error = 0.0
+                
+
+        def compute_errors(self, setpoint, position, previous_position, time_step):
+                self.position_error = setpoint - position
+
+
+                self.integral_error += self.position_error * time_step
+                self.integral_error = np.clip(self.integral_error, -self.I_MAX, self.I_MAX)
```
ParseDotEXE: if I_MAX is set to 0.0 this gives -> np.clip(self.intergral_error, -0.0, 0.0)
This caps the Integral error to exactly 0 -> could be handled differently maybe

ros2_ws/src/controls/include/controls/attitude_controller.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+                rclcpp::Subscription<imu_msg>::SharedPtr sub_imu_;
+                rclcpp::Subscription<geometry_msgs::msg::Quaternion>::SharedPtr sub_target_orientation_;
+
+                // State variables
+                quatd q_vi_;
+                Vec3 w_vi_;
+
+                //Target state variables
+                quatd q_vi2_;
```
ParseDotEXE: General comments for minor efficiency improvements:
1- Adding a quaternion validation check
2- This one is probably not that necessary but adding timeout handling for when the IMU data comes in choppy

ros2_ws/src/controls/include/controls/superimposer.hpp
```diff
@@ -0,0 +1,73 @@
...
+namespace controls
+{
+    using float64msg = std_msgs::msg::Float64;
+    using wrench_msg = geometry_msgs::msg::Wrench;
+    using imu_msg = sensor_msgs::msg::Imu;
+    using quatd = Eigen::Quaternion<double>;
+    using Mat3 = Eigen::Matrix3d;
+    using Vec3 = Eigen::Vector3d;
+
+    class superimposer: public rclcpp::Node
```
ParseDotEXE: follow class naming conventions
class name unclear of what the class does

ros2_ws/src/controls/launch/attitude_controller.launch.py
```diff
@@ -0,0 +1,29 @@
+# launch/thrust_mapper.launch.py
```
ParseDotEXE: does this launch attitude controller or thrust mapper?

ros2_ws/src/controls/scripts/depth_controller.py
```diff
@@ -0,0 +1,86 @@
...
+    def depth_callback(self, msg):
+        self.previous_depth = self.current_depth
+        self.current_depth = msg.data
+
+    def setpoint_callback(self, msg):
+        self.setpoint_depth = msg.data
+
+    def control_loop_callback(self):
+        # Compute PID errors
+        self.pid.compute_errors(self.setpoint_depth, self.current_depth, self.previous_depth, self.time_step)
```
ParseDotEXE: This and the next line could be merged into one

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,137 @@
...
+{
+    AttitudeController::AttitudeController(): Node("attitude_controller")
+    {
+        this->declare_parameter<double>("P_ex", 1.0);
+        this->declare_parameter<double>("P_ey", 1.0);
+        this->declare_parameter<double>("P_ez", 1.0);
+        this->declare_parameter<double>("P_wx", 1.0);
+        this->declare_parameter<double>("P_wy", 1.0);
+        this->declare_parameter<double>("P_wz", 1.0);
+        this->declare_parameter<double>("buoyancy", 278.0); // Newtons
```
ParseDotEXE: good unit declaration

build.sh
```diff
@@ -30,7 +30,7 @@ if [ -d "ros2_ws/src" ]; then
     EXCLUDE_PKGS+=(--packages-skip zed-ros2-wrapper zed_ros2 zed_components zed_msgs zed_wrapper)
   fi
   cd ros2_ws
-  rosdep install --from-paths src --ignore-src -r -y --skip-keys="zed zed_msgs zed_components"
+  # rosdep install --from-paths src --ignore-src -r -y --skip-keys="zed zed_msgs zed_components"
```
hutchallison: Can we add a comment for why this was removed, "overwrites other necessary dependencies"

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,137 @@
...
+        this->get_parameter("P_ey", P_ey_);
+        this->get_parameter("P_ez", P_ez_);
+        this->get_parameter("P_wx", P_wx_);
+        this->get_parameter("P_wy", P_wy_);
+        this->get_parameter("P_wz", P_wz_);
+        this->get_parameter("buoyancy", buoyancy_);
+        this->get_parameter("r_bv", r_bv_);
+        this->get_parameter("control_loop_hz", control_loop_hz_);
+
+        q_iv_ = quatd::Identity(); // Initial orientation: identity quaternion
```
ParseDotEXE: Uninitialized control before first IMU message - Control loop starts immediately but q_iv_ and w_iv_ only update on first IMU callback. Should either wait for first message or add validity check.

fix: wait for the first message

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,137 @@
...
+        );
+
+        control_timer_ = this->create_wall_timer(
+            std::chrono::milliseconds(static_cast<int64_t>(1000 / control_loop_hz_)),   // Control loop frequency
+            std::bind(&AttitudeController::control_loop_callback, this)
+            );
+    }
+
+    void AttitudeController::imu_callback(const imu_msg::SharedPtr msg)
+    {
```
ParseDotEXE: // In imu_callback
eg : imu_received_ = true;

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,137 @@
...
+        effort_msg.torque.z = total_torque.z();
+        effort_msg.force.x = 0.0;
+        effort_msg.force.y = 0.0;
+        effort_msg.force.z = 0.0;
+
+        return effort_msg;
+    }
+
+    void AttitudeController::control_loop_callback()
+    {
```
ParseDotEXE: // In control_loop_callback
eg: if (!imu_received_) return; // Skip until IMU available

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -0,0 +1,124 @@
+#include "controls/superimposer.hpp"
+
```
ParseDotEXE: include a documentation to provide what the class does, what the methods do and what are the params

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -0,0 +1,124 @@
...
+   }
+
+   void superimposer::imu_callback(const imu_msg::SharedPtr msg)
+   {
+       // Extract orientation quaternion from IMU message
+       q_iv_ = quatd(
+           msg->orientation.w,
+           msg->orientation.x,
+           msg->orientation.y,
+           msg->orientation.z
```
ParseDotEXE: missing normalization

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -0,0 +1,124 @@
...
+       );
+
+       sub_attitude_effort_ = this->create_subscription<wrench_msg>(
+           "/controls/attitude_effort",
+           1,
+           std::bind(&superimposer::attitude_effort_callback, this, std::placeholders::_1)
+       );
+
+       // Initialize publisher
+       pub_effort_ = this->create_publisher<wrench_msg>("/controls/total_effort", 1);
```
ParseDotEXE: README says output should be /controls/combined_effort?

ros2_ws/src/sensors/params/sensors_frames.yaml
hutchallison: Nicely Commented

.vscode/c_cpp_properties.json
JPGC04: Do not commit this file please

ros2_ws/src/bringup/launch/bringup.launch.py
JPGC04: Do we want a separate sim.launch file similarly to in auv-2025? This is not relevant to this PR specifically, but it is a point to discuss for when more sim stuff gets merged.

ros2_ws/src/controls/launch/depth_controller.launch.py
```diff
@@ -0,0 +1,23 @@
...
+
+def generate_launch_description():
+        # No Launch configuration variables
+
+        depth_controller = Node(
+                package='controls',
+                executable='depth_controller.py',
+                name='depth_controller',
+                output='screen',
+                parameters=[{
```
JPGC04: I don't know if I missed another file with those parameters, but would be nice to have KP KD and KI params for the sim cause the tuning will probably be a bit different in sim. Like you could have a sim condition for pids.

ros2_ws/src/propulsion/package.xml
```diff
@@ -12,8 +12,6 @@
   <url type="repository">https://github.com/mcgill-robotics/AUV-2026</url>
   <author email="nadim.asmar@mail.mcgill.ca">Nadim Asmar</author>
 
-<buildtool_depend>ament_python</buildtool_depend>
```
JPGC04: Is this package not required?

ros2_ws/src/sensors/launch/sensors.launch.py
JPGC04: The physical sensor launches in my opinion should be on their own, for example the display microros agent should be launched in the depth process launchfile. same for imu and eventually dvl.  The problem with this file is that if the display fails to launch, it could prevent the imu from starting up. so it would be better to seperate those in their own files, and then have sensors launch file to launch all those launch files. very similar to the auv-2025 architecture.

ros2_ws/src/sensors/params/sensors_frames.yaml
JPGC04: this is very nice and it looks like an amazing way to modify params out of launch files

build.sh
JPGC04: was it requested by vision people to remove this line?

JPGC04 (dismissed):
this pr is great and should be merged. I unfortunately don't have enough technical knowledge to comment on the specifics of the code. I would hope this is tested in sim before it gets merged, but ultimately this branch needs to be merged when possible.

ros2_ws/src/controls/package.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0"?>
+<package format="3">
+  <name>controls</name>
+  <version>0.1.0</version>
+  <description>Control nodes for the AUV.</description>
+
+  <maintainer email="dev@mcgillrobotics.com">McGill Robotics</maintainer>
+  <license>GPLv3</license>
```
ParseDotEXE: good

ros2_ws/src/controls/README.md
```diff
@@ -0,0 +1,121 @@
...
+    - [Subscribed Topics](#subscribed-topics)
+  - [Installation](#installation)
+    - [Dependencies](#dependencies)
+    - [Building](#building)
+    - [Running](#running)
+    - [License](#license)
+
+
+## Overview
+The controls package implements separate depth and attitude controllers. The depth controller uses a PID loop on `/processed/depth` and publishes a vertical effort on `/controls/depth_effort`. The attitude controller uses IMU orientation (`processed/imu`) and a target quaternion (`quaternion_setpoint`) to publish torques on `/controls/attitude_effort`. The superimposer node sums both efforts, applies optional bias terms, and publishes `/controls/combined_effort` for propulsion.
```
ParseDotEXE: great description

ParseDotEXE (commented):
done reviewing the PR! lmk if I missed anything or if you need anything to be cleared up

WilliamZhang2205 (changes requested):
Overall very nice stuff! In the future, we can add doc strings above each function to facilitate the code reader's understandability. While, it is mentioned in the code to refer to the README.md, nice docstrings can be shown on your screen when hovering over a function making comprehension easier and more accessible. That can be done in another PR later on. The only mandatory change I'd ask is the removing the .vscode folder as JP mentioned before.

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -0,0 +1,148 @@
...
+   }
+
+    void superimposer::y_effort_callback(const wrench_msg::SharedPtr msg)
+   {
+       y_effort_ = *msg; // This is in the pool frame
+   }
+
+   void superimposer::attitude_effort_callback(const wrench_msg::SharedPtr msg)
+   {
+       attitude_effort_ = *msg; // This already is in the AUV body frame
```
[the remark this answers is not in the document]
nadimasmar: This is a good idea. I will add this as an issue. New mech members will work on it. For this PR, I think we can let through the current version of superimposer.

However, what do you exactly mean by "reset". Do you mean to simply set the efforts to zero?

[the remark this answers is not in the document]
WilliamZhang2205: I mean we could have a service call equivalent to each effort callback. One service to set depth effort to 0, another to set x effort to 0, another for y effort to 0 and one last for attitude effort to 0.

WilliamZhang2205 (dismissed):

ros2_ws/src/controls/src/superimposer.cpp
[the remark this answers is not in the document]
nadimasmar: I think this can already be achieved by setting the `enabled` parameter to `false` for every respective controller.

Only difference is that its a `ros2 param set` instead of a `ros2 service`.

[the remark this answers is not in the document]
WilliamZhang2205: I still believe it should be a service or a feature to set it to 0. At least maybe for X and Y controllers. For example, say we had set the setpoint to x=5, y =2, Dougie gets there, and we turn the controller off. If we turn it on again after having moved Dougie to x=0, y=0,  Dougie could start moving immediately and hurt someone. A fix could be when you set the param to false, the code could set those respective efforts to 0.

JPGC04 (dismissed):

WilliamZhang2205 (dismissed):

WilliamZhang2205 (dismissed):

JPGC04 (dismissed):

ros2_ws/src/sensors/src/depth_processor.cpp
```diff
@@ -0,0 +1,102 @@
...
+            "/sensors/depth/z",
+            10,
+            std::bind(&DepthProcessor::depth_callback, this, std::placeholders::_1)
+        );
+
+    imu_sub_ = this->create_subscription<sensor_msgs::msg::Imu>(
+        "auv_frame/imu",
+        10,
+        std::bind(&DepthProcessor::imu_callback, this, std::placeholders::_1)
+    );
```
kaidalisohaib, line 27: For sensors subscribers, you should use rclcpp::SensorDataQoS() instead of 10 queue size to have the communication in "Best Effort" mode. 
Also, since the depth sensor is giving data at about 10 Hz (every 100ms), which is relatively slow, you would want to keep the queue size pretty small to keep the newest data. (rclcpp::SensorDataQoS().keep_last(2))
For imu, since it's faster it's ok to keep a queue size of 10, but you still want the best effort from the qos (rclcpp::SensorDataQoS().keep_last(10)).

ros2_ws/src/sensors/src/dvl_processor.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    this->declare_parameter<std::string>("frame_id_global", "pool_link");
+    this->get_parameter("frame_id_global", frame_id_global_);
+
+
+    imu_sub_ = this->create_subscription<sensor_msgs::msg::Imu>(
+        "auv_frame/imu", 10, std::bind(&DvlProcessor::imu_callback, this, std::placeholders::_1));
+
+    dvl_sub_ = this->create_subscription<nav_msgs::msg::Odometry>(
+        "dvl/odometry", 10, std::bind(&DvlProcessor::dvl_callback, this, std::placeholders::_1));
```
kaidalisohaib: Same idea for QoS of subscribers.
For imu rclcpp::SensorDataQoS().keep_last(10) is fine since it's very fast.
For dvl, 10 queue size at 5Hz (200 ms) is too much because it represents a potential delay of 2 seconds and since it's our first source of position we would want the latest data received. Therefore a queue size of 1 (rclcpp::SensorDataQoS().keep_last(1)).

ros2_ws/src/sensors/src/Imu_processor.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    imu_pub_ = this->create_publisher<imu_msg>(
+        "auv_frame/imu",
+        10
+    );
+    imu_sub_ = this->create_subscription<imu_msg>(
+        "imu/data",
+        10,
+        std::bind(&ImuProcessor::imu_callback, this, std::placeholders::_1)
+    );
```
kaidalisohaib, line 22: QoS update here as well.

ros2_ws/src/controls/scripts/depth_controller.py
```diff
@@ -0,0 +1,109 @@
...
+        qos = QoSProfile(
+            reliability=ReliabilityPolicy.RELIABLE,
+            durability=DurabilityPolicy.VOLATILE,
+            history=HistoryPolicy.KEEP_LAST,
+            depth=1,
+        )
+
+        self.pub_effort = self.create_publisher(Wrench, '/controls/depth_effort', qos)
+        self.sub_depth = self.create_subscription(Float64, 'auv_frame/depth', self.depth_callback, qos)
+        self.setpoint_sub = self.create_subscription(Float64, '/controls/depth_setpoint', self.setpoint_callback, qos)
```
kaidalisohaib, line 30: Same idea as the other QoS update comments. For the sub_depth and pub_effort, you should switch from Reliable to Best Effort (using `qos_profile_sensor_data`). In a controller, it is better to drop a lost packet and move to the next one than wait for a late, old update. 
For the setpoints it should stay reliable since we want to make sure to receive the goals sent compared to sensors data which happens many times per seconds.

ros2_ws/src/controls/scripts/planar_controller.py
```diff
@@ -0,0 +1,117 @@
...
+        qos = QoSProfile(
+            reliability=ReliabilityPolicy.RELIABLE,
+            durability=DurabilityPolicy.VOLATILE,
+            history=HistoryPolicy.KEEP_LAST,
+            depth=1,
+        )
+
+        self.pub_effort = self.create_publisher(Wrench, f'/controls/{axis_name}_effort', qos)
+        self.sub_position = self.create_subscription(PointStamped, f'auv_frame/dvl/position', self.position_callback, qos)
+        self.setpoint_sub = self.create_subscription(Float64, f'/controls/{axis_name}_setpoint', self.setpoint_callback, qos)
```
kaidalisohaib, line 33: Same idea as the other QoS comments. Except for setpoint which should stay at reliable.

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,188 @@
...
+        sub_imu_ = this->create_subscription<imu_msg>(
+            "auv_frame/imu",
+            1,
+            std::bind(&AttitudeController::imu_callback, this, std::placeholders::_1)
+        );
+        sub_target_orientation_ = this->create_subscription<geometry_msgs::msg::Quaternion>(
+            "/controls/quaternion_setpoint",
+            1,
+            std::bind(&AttitudeController::target_orientation_callback, this, std::placeholders::_1)
+        );
```
kaidalisohaib, line 53: For the IMU and effort topics, we should use rclcpp::SensorDataQoS() instead of a queue size of 1. Even though it seems counter-intuitive, Reliable mode (the default for 1) is actually slower because it pauses to retry lost packets. In a high-speed attitude loop, an old IMU reading is useless. It’s better to drop a lost packet and move immediately to the next one to keep the controller stable.

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -0,0 +1,188 @@
...
+        P_e_ << P_ex_, 0, 0,
+                0, P_ey_, 0,
+                0, 0, P_ez_;
+        
+        P_w_ << P_wx_, 0, 0,
+                0, P_wy_, 0,
+                0, 0, P_wz_;
+
+
+        pub_effort_ = this->create_publisher<wrench_msg>("/controls/attitude_effort", 1);
```
kaidalisohaib: For pub_effort, it's better to use Best Effort (rclcpp::SensorDataQoS()). If an effort topic is lost, we don't want the system to waste time retrying it. By the time the 'retry' arrives, it's already out of date. It's much safer for the thrusters to just receive the next fresh calculation immediately to keep the control loop smooth.

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -0,0 +1,160 @@
...
+       );
+
+       sub_attitude_effort_ = this->create_subscription<wrench_msg>(
+           "/controls/attitude_effort",
+           1,
+           std::bind(&superimposer::attitude_effort_callback, this, std::placeholders::_1)
+       );
+
+       // Initialize publisher
+       pub_effort_ = this->create_publisher<wrench_msg>("/controls/total_effort", 1);
```
kaidalisohaib: Same idea for QoS as the other comments.

ros2_ws/src/sensors/src/State_aggregator.cpp
```diff
@@ -0,0 +1,122 @@
...
+        dvl_velocity_sub_ = this->create_subscription<geometry_msgs::msg::TwistStamped>(
+            "auv_frame/dvl/velocity",
+            1,
+            std::bind(&State_aggregator::dvl_velocity_callback, this, std::placeholders::_1)
+        );
+
+        pose_pub_ = this->create_publisher<geometry_msgs::msg::PoseStamped>(
+            "state/pose",
+            1
+        );
```
kaidalisohaib, line 47: Same QoS updates here as well.

kaidalisohaib (changes requested):
For sensor subscribers, we should switch from Reliable to Best Effort using `rclcpp::SensorDataQoS()` or `from rclpy.qos import qos_profile_sensor_data` in python or a custom QoSProfile.
This profile is made specifically for sensors, it stops the system from retrying lost packets, which avoids lag. For our use case getting the newest data quickly is more important than making sure every single old packet arrives.

ros2_ws/src/controls/src/superimposer.cpp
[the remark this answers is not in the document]
nadimasmar: Say there is such a feature. When that feature is enabled, it overrides the controller logic and sets all efforts to zero. In the same scenario that you outline, disabling the feature would also cause the robot to jump suddenly, potentially hurting someone. 

if the `enabled` parameter is set to `false.` All controller efforts are set to zero and then published. Isn't that the same thing as what you are proposing?

ros2_ws/src/sensors/src/depth_processor.cpp
nadimasmar: Thank you for the suggestion. I implemented it. I set the queue size to be 1 for all topics. Once there is a newer reading, there is no value in getting the older one anymore.

ros2_ws/src/sensors/src/dvl_processor.cpp
nadimasmar: Implemented with queue size 1

ros2_ws/src/sensors/src/Imu_processor.cpp
nadimasmar: Implemented with queue size 1

ros2_ws/src/controls/src/attitude_controller.cpp
nadimasmar: Done

ros2_ws/src/controls/src/attitude_controller.cpp
nadimasmar: Done

ros2_ws/src/controls/scripts/depth_controller.py
nadimasmar: Done

ros2_ws/src/controls/scripts/planar_controller.py
nadimasmar: Done

ros2_ws/src/controls/src/superimposer.cpp
nadimasmar: Done

ros2_ws/src/sensors/src/State_aggregator.cpp
nadimasmar: Done

kaidalisohaib (approved):
Tested in the sim 👍 
Note: Hopefully simply tuning the controllers once again will fix the instability seen during pre-qual.

JPGC04 (approved):

PR #45, opened 2026-01-10
Unify Dev Environment: NVIDIA GPU/ZED Support & Universal CPU Fallback

kaidalisohaib:
Unifies the development environment by supporting high-performance NVIDIA GPU workflows and universal CPU-only fallbacks. Container images are now optimized to install only the dependencies compatible with the user's hardware.

## Major Changes
*   **Dual service (cpu/nvidia-gpu):** Refactored `docker-compose.yml` into `nvidia` and `cpu` services.
    *   `cpu`: Designed for machines without NVIDIA GPUs, excludes ZED SDK and CUDA to remain lightweight.
    *   `nvidia`: Designed for hardware-accelerated workflows, includes CUDA 12.8 and ZED SDK 5.1.1 to run perception stacks alongside the simulation.
*   **More flexible build system:** Rewrote `build.sh` with automatic hardware and SDK detection. it utilizes `AMENT_IGNORE` to selectively toggle ZED components, ensuring successful workspace compilation across all environments.
    * Adds `-c` `-d` `-o` and `-p` flag for building from scratch, debug build, offline build,,and building a specific package respectively. Specific configuration are explained in the [README](https://github.com/mcgill-robotics/AUV-2026/blob/ad1a37d0b5268f88c373c02a23f1a90707c9d336/README.md?plain=1#L81)
*   **Unified configuration:** Updated VS Code `devcontainer.json` to utilize Docker Compose as the "Single Source of Truth," eliminating image duplication and ensuring environment variables, volume mounts, and security settings are consistent between the CLI and VS Code.
    * Docker devcontainer configurations exist for both `cpu` and `nvidia`

## Minor Changes
* Adds `.colcon/default.yaml` file that sets `--symlink-install` as a default option when running `colcon build`
* Sets ZED SDK version to 5.1.1 in both Docker containers and zed-ros2-wrapper submodules
* Integrates urgent fix from #61

### DIFF
--- a/Docker/dev/Dockerfile.nvidia
+++ b/Docker/dev/Dockerfile.nvidia
@@ -50,6 +50,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         lsb-release \
         ca-certificates \
         sudo \
+        vim \
         # ZED SDK Dependencies
         zstd \
         libusb-1.0-0-dev \
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ Multiple build flags are available:
 -  `./build.sh` : Default Release Build, works for most cases
 -  `./build.sh -c` : Clean Build, removes previous build artifacts (`build/`, `install/`, `log/` folders)
 -  `./build.sh -d` : Debug Build, compiles with debug symbols and no optimizations and serial compilation
+-  `./build.sh -o` : Offline Build, prevents CMake from attempting to download missing dependencies
 -  `./build.sh -p <package_name>` : Build a specific package and its required dependencies (as determined by colcon)
 
 Since the debug `-d` build is single-threaded for easier debugging, it may take very long to compile. As such, it should only be used in tandem with the `-p` flag to build a specific package. A typical debug build command would be `./build.sh -cd -p <package_name>`. Note that the `-c` will only clean the build artifacts for the specified package and its dependencies.
--- a/Docker/dev/README.md
+++ b/Docker/dev/README.md
@@ -6,7 +6,7 @@ This directory contains the Docker setup for local development on x86_64 machine
 
 ### For Everyone
 *   **Docker Engine** & **Docker Compose**.
-*   **Disk Space:** Ensure you have at least **25GB** free (CUDA and PyTorch are large).
+*   **Disk Space:** Ensure you have at least **40GB** free (CUDA and PyTorch are large).
 
 ### For NVIDIA GPU Users
 If you have an NVIDIA GPU, you must install the **NVIDIA Container Toolkit** to allow Docker to access your hardware.


Docker/dev/unity_sim_zed_config.yaml
```diff
@@ -0,0 +1,24 @@
+/**:
+  ros__parameters:
+    general:
+      camera_model: 'zedx'
```
Rain1618: zed2i?

Docker/dev/docker-compose.yml
Rain1618: Just realised the reason we kept running into "Permission Denied" errors when editing files w/ vscode could be because we are mounting the project to /root/AUV-2026 so files created inside the container will be owned by root... (and the user is `jetson`) If this is still an issue, consider adding smth like `user: "${UID}:${GID}"`

avyan-k (commented):
Great changes overall, most of my comments where nitpicky suggestions and features requests that can easily be slotted into this PR. 

I can confirm the nvidia container works on a Windows machine with a NVIDIA GPU, the `zed-ros2-wrapper` can be built.

I also reviewed the changes of nvidia container compared to the [althack humble container](https://github.com/athackst/dockerfiles/blob/main/ros2/humble.Dockerfile) (since we are extending from the cuda container instead) and they have pretty much the same features

Alb33rt (changes requested):
Great changes. I got a few questions in the code. Please lmk why they are the case.

Furthermore, I'm curious as to why aren't we just using the `althack` cuda containers because they've been working well before? Do the nvidia containers just work better?

Docker/dev/Dockerfile
```diff
@@ -1,58 +1,122 @@
...
+RUN pip3 install --no-cache-dir \
+    ultralytics \
+    pandas \
+    scipy \
+    notebook \
+    roboflow \
+    albumentations \
+    typing_extensions \
+    sympy \
+    cython
```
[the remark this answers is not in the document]
avyan-k: Yes I think installing the packages should be optional, especially since unlike apt, simply running pip does not consider package versions and will override existing versions of the dependencies to match the package. This breaks the setup we have on the jetson right now (importantly opencv-contrib-python==4.12 and numpy<2.0 for the cv bridge to work) and even though this is the dev container, it should still have existing packages correspond to the jetson container.

I think it's better to have a `requirements.txt` with all the right package versions based on what's on the jetson (perhaps something to determine later for now the two above I mentioned), then we can just pip install them later if needed (maybe add an option to the `build.sh` but that perhaps puts too much burden on that script)

Docker/jetson/README.md
```diff
@@ -19,6 +19,25 @@ This repository contains a Dockerized ROS 2 Humble development environment for t
     docker exec -it jetson-douglas-1 bash
     ```
 
+### Network Configuration
+We standardize **`ROS_DOMAIN_ID=0`** across all our Docker containers. This is pre-configured in the `docker-compose.yml` environment variables. Ensure any external machines communicating with the Jetson also use `ROS_DOMAIN_ID=0`.
```
Alb33rt: Great addition. This is necessary for external devices to connect when running the same ROS network

Docker/dev/docker-compose.yml
```diff
@@ -0,0 +1,91 @@
...
+      - LIBGL_ALWAYS_SOFTWARE=1
+      # Added to match original devcontainer features:
+      - WAYLAND_DISPLAY=${WAYLAND_DISPLAY}
+      - XDG_RUNTIME_DIR=${XDG_RUNTIME_DIR}
+      - PULSE_SERVER=${PULSE_SERVER}
+
+    user: "douglas"
+    volumes:
+      - ../../:/home/douglas/AUV-2026
+      - ../../ros2_ws/src/vision/model_pipeline/data:/images
```
ArthurGuihaire: This mount technically isn't needed since AUV-2026 is already mounted so we can already write to model_pipeline, but we can keep it, its nice to have /images instead of a long path

Docker/dev/Dockerfile.nvidia
```diff
@@ -0,0 +1,267 @@
...
+ARG TENSORRT_MINOR=9
+
+# Set environment variables
+ENV LANG=C.UTF-8 \
+        LC_ALL=C.UTF-8 \
+        DEBIAN_FRONTEND=noninteractive \
+        ROS_DISTRO=humble \
+        # Critical for ZED SDK to access hardware
+        NVIDIA_DRIVER_CAPABILITIES=all \
+        TZ=Europe/Paris \
```
ArthurGuihaire: very minor but why not America/Montreal

README.md
```diff
@@ -26,6 +33,65 @@ If you have the software installed, follow these instructions:
...
+    source ros2_ws/install/setup.bash
+```
+
+Note the script can be run from anywhere in the container and will always build inside the `AUV-2026/ros2_ws` folder.
+
+Multiple build flags are available:
+-  `./build.sh` : Default Release Build, works for most cases
+-  `./build.sh -c` : Clean Build, removes previous build artifacts (`build/`, `install/`, `log/` folders)
+-  `./build.sh -d` : Debug Build, compiles with debug symbols and no optimizations and serial compilation
+-  `./build.sh -p <package_name>` : Build a specific package and its required dependencies (as determined by colcon)
```
ArthurGuihaire: Add documentation for offline flag

ArthurGuihaire (approved):

PR #55, opened 2026-01-11
Update CI to support manual monthly stable Docker releases

sarayangrh:
Previously, images were only pushed to Docker Hub when changes were merged to main.

This update allows us to manually release a stable version every month from a chosen branch. These releases will use a date tag (e.g., 2026-01-arm64).

### DIFF
--- a/.github/workflows/docker-images-main.yml
+++ b/.github/workflows/docker-images-main.yml
@@ -46,7 +46,7 @@ jobs:
         file: Docker/dev/Dockerfile
         platforms: linux/amd64
         push: ${{ github.event_name == 'workflow_dispatch' }}
-        tags: mcgillrobotics/auv_2026:${{ inputs.tag_name }}-amd64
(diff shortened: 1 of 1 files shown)


Rain1618 (dismissed):
A few general comments: It might be a good idea to check if there is an existing tag before running workflow_dispatch, rn I think that it will just overwrite it on Docker Hub without warning... Also for like general knowledge purposes, to avoid repeated code you can use [this]( https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations). Looks good tho, I'm gonna approve and up to you if you wanna implement changes!!

Alb33rt:
Could we potentially have a cron job that does this manually automatically? Or would you rather manually push it up every month.

Alb33rt (dismissed):
LMK if you prefer to do this manually or with a cron job.

Alb33rt (approved):
I like the fact that you are using a matrix to run the different strategies rather than manually defining multiple processes. Great work!

hutchallison (approved):

PR #61, opened 2026-02-01
Minor but Urgent fix: Make jetson Docker container source appropriate ros2 setup.bash

avyan-k:
## Issue
* Sourcing the container from `/opt/ros/humble/install/setup.bash⁩` prevents the colcon from finding all required package (notably `rosidl_generator_cpp` to generate custom services and messages). 
* `/opt/ros/humble/install/⁩` is used to build ROS itself while `/opt/ros/humble/⁩` the packages are installed, but since we source from `/opt/ros/humble/install` [in this line of the container](https://github.com/mcgill-robotics/AUV-2026/commit/807948867b6d6a2f838928eb811a552e7b5d073f#diff-f20543f1c0618a14daec98c46e200aca01595286db19c7e5c27a82ac6176c191L236), `colcon build` is unable to find `rosidl_generator_cpp` and fails.

## Fix
Replace `source /opt/ros/humble/install/setup.bash` with `source /opt/ros/humble/setup.bash` in the `.bashrc` so that *only* `/opt/ros/humble` is sourced. Unsure why this issue only pops up now but this will be tested. For now we add this fix to allow the jetson build ROS packages on the `main` branch

### DIFF
--- a/Docker/jetson/Dockerfile
+++ b/Docker/jetson/Dockerfile
@@ -233,5 +233,5 @@ RUN pip install --extra-index-url ${PIP_INDEX_FOR_NON_NVIDIA} notebook \
 	pip install --ignore-install --upgrade typing_extensions
 
 # Source env variables
-RUN echo "source /opt/ros/humble/install/setup.bash && source /root/dependencies_ws/install/setup.bash && source /root/AUV-2026/ros2_ws/install/setup.bash" >> ~/.bashrc
+RUN echo "source /opt/ros/humble/setup.bash && source /root/dependencies_ws/install/setup.bash && source /root/AUV-2026/ros2_ws/install/setup.bash" >> ~/.bashrc
 CMD ["bash"]
--- a/Docker/jetson/README.md
+++ b/Docker/jetson/README.md
@@ -19,6 +19,22 @@ This repository contains a Dockerized ROS 2 Humble development environment for t
     docker exec -it jetson-douglas-1 bash
     ```
 
+## ⚠️ Warnings and pitfalls
+
+- When build the ROS package, make sure to source the ROS setup files first:
+    ```bash
+    source /opt/ros/humble/setup.bash
+    source /root/dependencies_ws/install/setup.bash
+    source /root/AUV-2026/ros2_ws/install/setup.bash
+    ```
+This is done by default when launching the container or opening a new bash shell. 
+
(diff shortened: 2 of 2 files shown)


WilliamZhang2205 (dismissed):
Needed to not cause rosidl generator errors when colcon building for micro_ros_msgs

asiancooki (changes requested):
Would this affect sourcing install/setup.bash after colcon build?

Do we still need to --symlink-install? If so, can you just add that to the colon defaults.yaml config file instead for ease of use?

JPGC04 (approved):
Isn't this how the environment should be sourced anyways.
You are supposed to be sourcing /opt/ros/humble/setup.bash not the install one

What am I missing?

WilliamZhang2205:
> Isn't this how the environment should be sourced anyways. You are supposed to be sourcing /opt/ros/humble/setup.bash not the install one
> 
> What am I missing?

The issue stems from dustynv's ros2 manual ros2 install and the ros2 we get sudo apt-get installs. Whenever we sourced /opt/ros/humble/install/setup.bash we would source dusty's manual install which has a bunch of manual package installs that are actually needed for zed2i's ros2 dependencies packages install. However, when we sudo apt-get install a ros-humble package, apt thinks we do not have a ros2 installation since the apt ros version is indeed not installed. So it installs another ros2 humble on the side. This is the /opt/ros/humble/setup.bash We confirmed this by running a container on the dustynv's base image and verifying that there was no /opt/ros/humble/setup.bash until we ran a apt-get install ros-humble-package, while the base image did contain a /opt/ros/humble/install/setup.bash 

Now to the build issues for micro_ros_msgs and auv_msgs. These packages rely on the rosidl_generator_cpp package to make msgs. However, dustynv's manual install has an outdated rosidl version. Back in september, I chalked the discrepancies of there being two setup.bash to one being the base ros2 sourcing and the install/setup.bash being our apt-get packages sourcing. Hence why the last lines of the dockerfile sourced opt/ros/humble/install/setup.bash

The fix we do is source dustynv's ros2 for building the dependencies needed for zed2i packages. This is the part of the dockerfile where there is a dependencies_ws folder. For our development purposes, we need the up to date rosidl_generator_cpp, so we source only the apt ros2 one. This however begs the question to how the "wrong" setup we did for months worked for a while and whether there are any of dustynv's ros2 packages that may be custom CUDA, which could optimize packages. 

This fix is also included in Sohaib's branch where there are more docker features (which may actually be merged tmr, once Avyan checks that all features of that branch do run on the actual jetson tmr). The purpose of this PR here is to roll out a working version first for you guys (if tmr's test fails) while we would investigate those questions and leads from the above paragraph and iron out the rest of the docker features on Sohaib's branch.

Alb33rt (approved):
Change is required because the `dustynv` container isn't updated to the latest version of `rosidl_generator`. This means that we manually build the package, which ends up in `/opt/ros/humble/setup.bash`. This logic makes sense and should be a standard from now on.

PR #63, opened 2026-02-03
Dev/foxglove integration

kaidalisohaib:
### WIP: Foxglove Studio Integration & Telemetry Infrastructure

**Summary**
Resolves #49  This PR establishes the foundation for AUV telemetry and remote monitoring. It introduces a central package for topic configuration and provides initial layout templates for Foxglove Studio.

> [!NOTE]
> This is a Work in Progress. Topic aliases and panel layouts are starting points and will change as our vision of the final design gets refined. 

**Key Components**
*   **Telemetry Package**:
    *   Centralized topic configuration file using semantic aliases.
    *   Launch configuration for the Foxglove Bridge (WebSocket server).
    *   Initial layout templates for:
        *   **Mission Control**: High-level status and cameras.
        *   **Sensors**: Low-level hardware validation.
        *   **Controls**: PID tuning and thruster feedback.
        *   **Teleoperation**: Manual control interface.
        *   **Perception**: Vision system and YOLO debugging.
        * **Electrical Layout**: System health monitoring
*   **Vision Integration**:
    *   Expanded image collection service to support down-camera capture toggling.

**Testing Instructions**
1.  Build the new package: `colcon build --packages-select telemetry`
2.  Launch the bridge: `ros2 launch telemetry dashboard.launch.py`
3.  In Foxglove Studio, connect to `ws://localhost:8765` and import the desired layout from the `foxglove` directory.

PR #99, opened 2026-04-09
Integrate DVLa50 Sensor Driver

### LINKED ISSUE #41: Integrate DVL
## 🎯 Objective
Get DVL data

## 📖 Description
If elec ethernet works, we need to tes [[Package ]](https://github.com/ndahn/dvl_a50) (already integrated wé docker)
If elec ethernet fails, implement serial parsing 

## ✅ Acceptance Criteria
- [ ] 

## 🔗 Links
**Related:** [#issue], [#issue]
**Blocks:** [#issue]
**Blocked by:** [#issue]

### LINKED ISSUE #97: Resolve DVL connection issues and document process [WIP]
Connection the DVL to the jetson via USB-ethernet seems to cause problem, where occasionally DVL cannot be pinged on boot. Try to reproduce the issue and resolve and document the process. Be sure to include tests and troubleshooting in the AUV 2026 wiki

avyan-k:
Resolves #41 and #97. This PR introduces support for the Water Linked DVL A50 device, supporting both Ethernet (JSON TCP) and Serial connections. It adds udev/network configuration scripts, refactors the `sensors` package to match the updated `dvl_msgs` types, and integrates a highly robust Python serial driver in place of the deprecated C++ serial driver.

> [!NOTE]
> The current recommended and most reliable driver is the **Serial Python driver (`dvl_a50_serial_python`)**, which has been fully enhanced to support strict CRC-8 checksumming, robust early-wakeup synchronization, and live product/firmware detail querying.

### 1. DVL Device Drivers & Submodule Changes
- **[NEW] Added [`dvl_a50_serial_python`](https://github.com/mcgill-robotics/dvl_a50_serial_python)** as a proper Git Submodule under `/ros2_ws/src/dvl_a50_serial_python`. This Python node serves as the primary serial driver.
- **[REMOVED] Removed C++ driver submodule [`dvl_a50_serial`](https://github.com/mcgill-robotics/dvl_a50_serial)** from tracking, avoiding duplicate drivers and simplifying compilation.
- **Added [`dvl_a50`](https://github.com/mcgill-robotics/dvl_a50)** (Ethernet driver) as a submodule.
- **Added [`dvl_msgs`](https://github.com/mcgill-robotics/dvl_msgs)** containing custom DVL ROS messages and services.
- **Removed redundant messages** in `auv_msgs` (`DeadReckonReport.msg` and `VelocityReport.msg`) which have been superseded by `dvl_msgs`.

### 2. Python Serial Driver Enhancements (`dvl_a50_serial_python`)
The Python driver has been fully optimized to align with strict serial protocol standards:
- **Strict CRC-8 Checksumming**: Automatically calculates and appends NMEA-style checksums (`*{crc:02x}\r\n`) to all outgoing serial configuration commands, preventing hardware NAKs.
- **Robust Early-Wakeup Response Synchronization**:
  - Implemented a thread-safe `command_event` and `wait_for_ack` flag state machine.
  - The node now terminates its block-wait immediately when receiving a valid configuration ACK (`wra` or `wrc`), or failure messages (NAK `wrn`, malformed request `wr?`, or checksum error `wr!`), rather than waiting for the entire timeout duration.
  - Spurious line/umbilical noise received while no command is active is filtered and ignored.
  - **Result**: Reduced startup configuration sequence latency from multiple seconds of timeouts down to **0.39 seconds**.
- **Product Details & Resilient Version Parsing**:
  - Sourced the product hardware details query (`wcw` / `wrw`) at startup to parse product name, version, chip ID, and optional IP address.
  - Made the firmware version report parser (`parse_wrv`) resilient against multiple formats, supporting dot-separated (`wrv,2.6.0`), comma-separated (`wrv,2,6,0`), and part-number embedded versions.

### 3. Jetson Network & Uudev Configuration
- Added the `jetson_scripts` folder containing connection and bridging scripts:
  - `create_usb_nmcli_connection.sh`: Configures network connections for DVL Ethernet connections using `nmcli`.
  - `createBridge.sh`: Bridges the Jetson and DVL network interfaces (should be used with caution as it can impact SSH connections).
  - `detect_device.sh`: Identifies the device path of USB drivers dynamically.
  - `generate_udev_rules.sh`: Generates standard rules for USB devices.
- **Udev Rule Update**: Updated `dvl.rules` to correctly match the active vehicle serial ID (`DU0DMHVL`) to symlink the DVL device to `/dev/dvl`.

### 4. Build Script & Sensors Package Refactoring
- Updated the build script offline option to bypass `git submodule update --init`, allowing offline live testing of local submodule edits on the Jetson without requiring parent commits.
- **Sensors Package Refactoring**: Refactored `dvl_processor.cpp` to subscribe directly to `dvl/velocity` and `dvl/dead_reckoning` topics from `dvl_msgs` rather than `dvl/odometry`.
  1. **Sensor Health Integrity**: Allows checking the DVL's health flags (`velocity_valid` and `status`) to handle faulty output upstream.
  2. **Real-Time Responsiveness**: Avoids driver-side buffering/blocking (the serial driver waits for slower dead reckoning packets, and the Ethernet driver buffers both velocity and dead reckoning before publishing `dvl/odometry`). Subscribing directly yields real-time, zero-lag DVL telemetry.

#### Ethernet
- Ethernet connection has access to the web GUI, which is much more feature rich that either the JSON ethernet or serial protocols, including notably
   - an visual indication of the DVL's pose and covariance 
image
[[1]](https://docs.waterlinked.com/img/dvl_gui_dashboard.png)

   - Graphs of transducer signal strength over distance and signal density for all frequencies
image
[[2]](https://dvl.demo.waterlinked.com/#/diagnostic)

   - Estimated temperature of the DVL and option to flash to verify connection
image
[[3]](https://dvl.demo.waterlinked.com/#/diagnostic)

DVL pose and configuration can be reproduced with driver only using Foxglove #63. But signal graphs, temperature and flash is not available.

- A limited simulation of the JSON protocol is provided in a dev branch of [AUV Unity Simulation](https://github.com/mcgill-robotics/auv-sim-unity/tree/dev/DVLTCPServer). While it provided accurate simulation data, it does not properly implement service responses and instead simply logs them for now.

#### Serial
- Significantly simpler electrical setup
- Less connection/boot issues that come with ethernet to serial connection (does not really happen with switch but the switch is not an ideal long term option)

## Annex
Commit before sensors change a606148a93d59393d283fe404d32d715b4285429
[output.webm](https://github.com/user-attachments/assets/f7a2f900-e175-4c57-9e78-5540a68d8da1)

Commit after with same setup ef9b7ea821939b560f34f024ff0b8999ebadd0d4
[output2.webm](https://github.com/user-attachments/assets/e35703c9-eb39-49e9-a63a-84abd04fb86c)


WilliamZhang2205 (dismissed):
lgtmmmm 👍

asiancooki (changes requested):
Overall this looks solid, and the PR description does a good job explaining the driver tradeoffs and why sensors now subscribes directly to dvl/velocity and dvl/dead_reckoning instead of relying on dvl/odometry.

I agree that the serial path seems like the practical default for now given the simpler setup and fewer connection issues (much easier on elec side).

My main feedback is mostly on documentation / merge readiness:

1. Please make the recommended default path even more explicit near the top of the PR. Right now I can infer that serial is the preferred working setup, but I think it would help to clearly state something like: “Use serial by default; ethernet support is included but experimental / secondary for now.” (just a suggestion)
2. I think the PR would benefit from a short “happy path” setup section for someone bringing this up from scratch on the Jetson (like a readme or smtg). For example:

- init/update submodules
- any Docker/build requirements
- udev / ModemManager steps
- which launch file to use for the serial driver
- expected topics to verify after bringup

3. For the network scripts, I’d add a stronger warning around createBridge.sh since the PR notes that it can break SSH. It would be good to mark it clearly as not part of the recommended setup unless specifically testing bridge mode.
4. Since this PR introduces multiple DVL-related submodules/packages, I’d also like one quick note confirming the intended long-term package set. The commit history mentions some message-package stuff, so a short final statement in the PR description about what is now actually required would make future maintenance easier.
5. Before merge, the failing Jetson / sensors checks should definitely be resolved, since that makes it hard to tell whether the branch is in a reviewable merge-ready state yet.

Other than that, nice and keep up the good work!

avyan-k:
> Overall this looks solid, and the PR description does a good job explaining the driver tradeoffs and why sensors now subscribes directly to dvl/velocity and dvl/dead_reckoning instead of relying on dvl/odometry.
> 
> I agree that the serial path seems like the practical default for now given the simpler setup and fewer connection issues (much easier on elec side).
> 
> My main feedback is mostly on documentation / merge readiness:
> 
>     1. Please make the recommended default path even more explicit near the top of the PR. Right now I can infer that serial is the preferred working setup, but I think it would help to clearly state something like: “Use serial by default; ethernet support is included but experimental / secondary for now.” (just a suggestion)
> 
>     2. I think the PR would benefit from a short “happy path” setup section for someone bringing this up from scratch on the Jetson (like a readme or smtg). For example:
> 
> 
>     * init/update submodules
> 
>     * any Docker/build requirements
> 
>     * udev / ModemManager steps
> 
>     * which launch file to use for the serial driver
> 
>     * expected topics to verify after bringup
> 
> 
>     3. For the network scripts, I’d add a stronger warning around createBridge.sh since the PR notes that it can break SSH. It would be good to mark it clearly as not part of the recommended setup unless specifically testing bridge mode.
> 
>     4. Since this PR introduces multiple DVL-related submodules/packages, I’d also like one quick note confirming the intended long-term package set. The commit history mentions some message-package stuff, so a short final statement in the PR description about what is now actually required would make future maintenance easier.
> 
>     5. Before merge, the failing Jetson / sensors checks should definitely be resolved, since that makes it hard to tell whether the branch is in a reviewable merge-ready state yet.
> 
> 
> Other than that, nice and keep up the good work!

Thank you for the review, many fair points! I have addressed most of the requested changes, here is a summary and some questions:

1. and 2. The driver is meant to be launched by the sensors package, but due to issues with the driver, it was only recently added as to not impact sensor output. I realized after your review that it should probably not always be launched since we cannot always guarantee the DVL is underwater, and so by default it is not launched, I have added extra instructions in the sensors README for launching the package with the DVL. However, I am not 100% sure if this should be the workflow and so I'd appreciate some feedback on this

I have also added a README.md to the jetson scripts folder to better describe why they are there and how to use them


3. I added a warning to the jetson script README and mentioned earlier and a warning to the script itself. It will now ask for confirmation (which default to no) before building the bridge

4. `dvl_a50_serial` is the main driver we will use, used to parse the serial protocol. `dvl_a50` is for the Ethernet protocol and I believe it should be kept as a backup. The `dvl_msgs` submodule is a generic message package (similar to [`auv_msgs`](https://github.com/mcgill-robotics/AUV-2026/tree/dev/dvl/ros2_ws/src/auv_msgs)) that defines messages specific to the DVL Waterlinked Protocol, and while `auv_msgs` did previously define the [`DeadReckonReport.msg`](https://github.com/mcgill-robotics/AUV-2026/blob/bb8239a8a4dd24ffa5e7615bbfb6c96270d97ae7/ros2_ws/src/auv_msgs/msg/DeadReckonReport.msg) and  [`VelocityReport.msg`](https://github.com/mcgill-robotics/AUV-2026/blob/bb8239a8a4dd24ffa5e7615bbfb6c96270d97ae7/ros2_ws/src/auv_msgs/msg/VelocityReport.msg) message type, our fork of the dvl_msgs repository more closely matches the protocol and it is overall more flexible to keep it separate along with the driver. Another option would be to add the Docker container similarly to the micro ros agent package but any time we need to update this message type it would need to updated in the Dockerfile which boils down to the same thing as just pinning a submodule.

5. The sensors package failing to build comes from a legacy CI  back when we were not sure which ROS packages/drivers the sensors package was dependent on and thus the CI would build it in an isolated manner. This is no longer possible, as with the update to sensors, `dvl_msgs` is a direct build dependency of the package Honestly in my opinion this CI workflow should eventually be removed entirely as we already have a workflow that build both the Docker container and ROS packages, thereby making this one redundant. For now I have added a "band-aid" fix but a proper CI update is being worked on another branch.

asiancooki (dismissed):
The added docs in the sensors README and jetson scripts README help a lot, and the stronger warning around createBridge.sh is definitely good to have. The clarification around the package split also makes the long term setup much easier to understand.

For the workflow question, I think not launching the DVL by default is the right call. We cannot always assume the DVL is connected, configured, or in a state where we actually want it running, especially if it is not underwater, so making it opt-in through the sensors launch flow seems safer and more explicit.

I think the important part is just that the intended workflow is very obvious in the docs:

- normal sensors bringup does not start DVL by default
- DVL can be enabled explicitly when needed
- the recommended path is the serial driver

The CI explanation also makes sense to me for this PR. Since the checks are passing now and the main documentation concerns were addressed, this looks good to me now.

Nice work, approving.

nadimasmar (dismissed):
The ROS interface of the dvl processor is correctly modified to accomodate the new message types. Thank you for verifying that the output doesn't change.

Approved

asiancooki (dismissed):
Thanks for the follow-up, I think this makes sense.

I still like the idea of not launching the DVL by default from sensors.launch.py directly, but having bringup.launch.py turn it on is fair if bringup is supposed to be the full mission-ready script.

At that point the assumption is already that the person launching bringup knows they are bringing up the whole robot and all the related risks, not just the DVL but also propulsion and controls.

The README warning makes that a lot clearer too, so I think the behavior is documented well enough now.

Also good catch on the sensors launch bug fix.

This looks good to me now, approving.

kaidalisohaib (approved):
Has been working for all the previous pool tests we've had and nothing broke

WilliamZhang2205 (approved):

asiancooki (approved):
Well done. Merge it.

PR #60, opened 2026-01-18
Object Map

avyan-k:
Resolves #31 and #46. Uses YOLO model from #67 and object detection nodes from #83. Will use sensor data from DVL #71 and IMU/Depth Sensor #43

## Main Changes
* Added Object Map ROS2 C++ Node
    - the node uses the ZED SDK directly instead of the [zed-ros2-wrapper](https://github.com/stereolabs/zed-ros2-wrapper), as the wrapper was found to have latency issues (see #62)
* Added `zed_detection.hpp` that collects all ZED SDK logic. This makes it easy to refactor out the ZED SDK if we decide to use the wrapper instead
* Added a compilation definition (see [`HAS_ZED_SDK`](https://github.com/mcgill-robotics/AUV-2026/blob/1c4e632014d28f6edc3c67b2011f546362de6af4/ros2_ws/src/vision/CMakeLists.txt#L66)) to make the inclusion of the ZED SDK is optional. This should allow building of the vision package to work on machines without NVIDIA GPUs or CUDA

* Refactored `vision` package to support both C++ and Python nodes
* Updated zed-ros2-wrapper to ROS2 Humble 5.1 to match with ZED SDK
* A PR following this one will eventually use DVL IMU and Depth sensor data to determine 3D positions of objects, uses measurements from ZED SDK (notably VIO) as a proxy

## Tests Done
* Test building on the jetson and see if ZED SDK is able to connect
* Determining necessary dependencies on the jetson (update Dockerfiles if needed)

## Minor Changes
* Changed VS Code editor default tab space to 4
* Updated `VisionObject` from #30 
* Updated the `VisionObject.msg` definition to include an `id` field for improved object identification.

PR #103, opened 2026-05-02
Refactor vision: monolithic tf2 pipeline, pyzed integration, and foxglove enhancements

### LINKED ISSUE #57: Train YOLOv8n and performance on down-cam to verify performance
## 🎯 Objective
Train a Robosub dataset captured from competition on a `YOLOv8n` model and infer on the down-cam. This should be done for both the real-life dataset captured from competition, but also for a synthetic dataset captured from a simulation. 

## 📖 Description
Train the `YOLOv8n` model, then perform inference and verify the performance. Document the findings both using diagrams (loss functions for example), and your findings. Re-train the model and document what changes you have made to make sure that this worked.

## ✅ Acceptance Criteria
- [ ] `YOLOv8n` **.pt** file trained
- [ ] Verify both the **mAP50** and **IOU** values within simulation
- [ ] Verify in real-life camera during a pooltest

## 🛠️ Implementation
**Approach:** Start by just training a model and prototyping a really lightweight inference node that would downsample the downcam resolution to the YOLO model resolution. Add some inference on this and document.
**Dependencies:** 

## 🔗 Links
**Related:** [#issue], [#issue]
**Blocks:** [#issue]
**Blocked by:** [#issue]

kaidalisohaib:
## Description
This PR introduces a major architectural refactor to the vision package, migrating to a monolithic `tf2_ros` workflow with synchronous detection, while enhancing object tracking, native ZED integration, and Foxglove visualization. 

### Key Features & Architecture
- **Monolithic TF2 Pipeline**: Replaced legacy matrix projections with native `tf2_ros` world-frame lookups for synchronous object detection.
- **Synchronous Detection**: Introduced the `VisionDetectionFrame` message to bundle exact AUV odometry snapshots with bounding box data, ensuring spatial consistency.
- **PyZED Native Ingestion**: Refactored image ingestion to process raw 3D points directly using the PyZED API, removing the reliance on `zed-wrapper`.
- **VIO Separation**: Separated vision VIO from the official pose TF published by `state_aggregator`, adding an `enable_vio` toggle to cleanly disable VIO when necessary.

### Perception & Object Map Enhancements
- **Memory Tracking**: Added `frames_since_last_seen` flag to `VisionObject` messages to differentiate between actively seen objects and those tracked from memory.
- **Object Publish Modes**: Replaced basic XY inheritance for octagons with explicit publish-time modes (e.g., table-from-octagon, table/octagon midpoint output modes).
- **YOLO Pipeline Improvements**: Added logic for minimum separation distances, depth confidence thresholds, and same-label-only track spawning.
- **Object Map Filtering**: Implemented border exclusion for the object map prior to ingesting into the ZED SDK.

### Telemetry & Visualization
- **Foxglove Enhancements**: Improved the front camera visualization by publishing ZED calibration info, adding an optical frame, and keeping 3D detections in the appropriate frame.
- **Simplified Feeds**: Consolidated the front camera feed to a single topic with a toggle for raw vs. annotated images.
- **Image Collection**: Restored `AutomaticCapture` interval support for dataset collection directly within the detection node.
- **Bringup Integration**: Added telemetry to the main `bringup.launch.py`.

### System & Tooling Updates
- **Dependencies**: Built native `CustomObjectDetectionRuntimeParameters` bindings, added ONNX/TRT configurations, and updated Jetson/dev Docker dependencies.
- **Sync**: Propagated `use_sim_time` through sensor launches to ensure the TF tree stays in sync.
- **Documentation**: Updated `README.md` and configuration docs to reflect the new pipeline and training scripts.

Merge #60 before this one.


ros2_ws/src/vision/model_pipeline/inference_test.py
```diff
@@ -0,0 +1,68 @@
...
+    image = cv2.imread(image_path)
+    
+    start_time = time.time()
+    predictions = model(image, confidence=0.4)
+    total_inference_time += (time.time() - start_time)
+
+    # Convert to supervision and visualize
+    detections = predictions[0].to_supervision()
+    
+    # labels = [
```
WilliamZhang2205: Shouldn't these be uncommented for the merge?

ros2_ws/src/vision/models/export_rfdetr.py
```diff
@@ -0,0 +1,4 @@
+from rfdetr import RFDETRSmall
+
+model = RFDETRSmall(pretrain_weights="robosub2026-synthetic-rfdetr_s.pth")
+model.export()
```
WilliamZhang2205, line 4: maybe rename this to export rdfetr

Docker/dev/Dockerfile
```diff
@@ -4,6 +4,10 @@
 # ---------------------------------------------------------
 FROM althack/ros2:humble-dev
 
+# Get uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```
avyan-k: Pin to specific version, I believe 0.11.9 is the latest but good to check, then replace this line with
```Dockerfile
COPY --from=ghcr.io/astral-sh/uv:0.11.9 /uv /uvx /bin/
```
as seen in [documentation](https://docs.astral.sh/uv/guides/integration/docker/#installing-uv)

Docker/dev/Dockerfile.nvidia
```diff
@@ -14,6 +14,15 @@ ARG UBUNTU_RELEASE=22
 # FROM ${BASE_IMAGE}
 FROM tuttelikz/opencv-cuda:4.10.0-cuda12.6.3-arch8.9-ubuntu22.04
 
+# Get uv
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
```
avyan-k: Pin version as with cpu

Docker/dev/Dockerfile.nvidia
```diff
@@ -76,6 +85,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         # GUI Support (RQT/Rviz)
         libgl1-mesa-glx \
         libgl1-mesa-dri \
+        # Tensorrt for model export
```
avyan-k: Is tensor rt installed here or somewhere else or at all? Lowkey confusing comment

Docker/jetson/Dockerfile
```diff
@@ -4,6 +4,9 @@ ARG IMAGE_NAME=dustynv/ros:humble-desktop-l4t-r36.4.0
 
 FROM ${IMAGE_NAME}
 
+# Get uv
```
avyan-k, line 7: Pin version as with cpu

Docker/jetson/Dockerfile
```diff
@@ -361,7 +348,8 @@ WORKDIR /home/$USERNAME/AUV-2026
 
 # Source env variables
 # We use /opt/dependencies_ws which is readable by all
-RUN echo "source /opt/ros/humble/setup.bash && source /opt/dependencies_ws/install/local_setup.bash && source /opt/ros/micro_ros_ws/install/local_setup.bash && source /opt/ros/foxglove_ws/install/local_setup.bash && source /home/$USERNAME/AUV-2026/ros2_ws/install/setup.bash" >> ~/.bashrc
```
avyan-k, line 364: Apparently the merge install option allows us to not have to source at all ,which in some cases may save in memory when loading the image. I believe micro ros already merge installs so we should be able to test removing that line from here. I'll look into it a bit more, and test it out, would be good to do it here since we are cleaning up the dockerfile anyway

Docker/jetson/down_cam.rules
```diff
@@ -1 +1 @@
-SUBSYSTEM=="video4linux", KERNEL=="video[0-9]*", ATTRS{idVendor}=="32e4", ATTRS{idProduct}=="9520", ATTRS{serial}=="3610000.usb", SYMLINK+="down_cam"
```
avyan-k, line 1: just to be sure you guys confirmed that copying these rules into the `/etc/udev/rules.d/` folder and symlinking works right? NOt for this PR but it might be helpful to mount these rules directly into the docker containers dev space so we can track changes. I might add it to #99 to include dvl rules. But please confirm they are the same

ros2_ws/src/sensors/src/State_aggregator.cpp
```diff
@@ -108,7 +109,18 @@ namespace sensors
         pose_msg.pose.position.z = -current_depth_; // Down is negative Z
 
         pose_pub_->publish(pose_msg);
-}
+
+        if (publish_pose_tf_) {
```
avyan-k, line 113: just for my own understanding why is this necessary versus the regular sensors output? Is it for foxglove visualization?

ros2_ws/src/sensors/params/sensors_frames.yaml
```diff
@@ -4,6 +4,7 @@
     frame_id_auv: "auv_link"
     frame_id_global: "pool_link" # Inertial frame, located at AUV's deployment point. 
     # Orientation is that of auv_link at deployment time. 
+    publish_pose_tf: true
```
avyan-k, line 7: If it is indeed for foxglove visualization, it might be good to have it false by default into the actual merge to main so we do not have pointless computations running in the actual mission. Maybe have a debug mode in bringup for this purpose similar to sim

ros2_ws/src/telemetry/README.md
avyan-k, line 1: Revert this change since it was from the telemetry merge

avyan-k (changes requested):
I haven't reviewed vision package changes yet which is the meat of the PR (I will do this soon), but since you guys made it work on the issues with no major issues, I do not expect having to request too many changes for them. I will post a full review later, this is just to have the early changes flagged

Docker/jetson/Dockerfile
```diff
@@ -184,62 +196,32 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
-    wget https://github.com/ros-perception/image_common/archive/refs/tags/${IMAGE_COMMON_VERSION}.tar.gz -O - | tar -xvz && mv image_common-${IMAGE_COMMON_VERSION} image_common && \
-    # copy needed packages to src
-    cp -r geographic-info/geographic_msgs/ . && \
-    cp -r mavros/mavros_msgs/ . && \
-    # clean up unneeded source folders
-    rm -rf geographic-info && \
-    rm -rf mavros && \
-    rm -rf /opt/dependencies_ws/src/.cache
-
 # ROS packages that do not depend on OpenCV/that ZED wrapper doesn't depend on can be installed here
```
WilliamZhang2205: Remove the zed wrapper comment

Docker/jetson/requirements.txt
```diff
@@ -0,0 +1,28 @@
...
+sympy
+cython
+setuptools<70
+typing_extensions
+
+torch 
+torchvision
+
+# Main Libraries
+ultralytics
```
WilliamZhang2205, line 11: For safety, we should pin all the versions here, we never know when something might update and kill CI. Add the pins in the constraints.txt. To see all the current versions, you guys could go on jetson and pip show to extract all the currently installed versions

ros2_ws/src/sensors/src/State_aggregator.cpp
WilliamZhang2205: If so add a comment, it can be confusing for anyone going through this the first time.

ros2_ws/src/telemetry/README.md
```diff
@@ -75,8 +75,6 @@ ros2 service call /image_collection/toggle_front_collection auv_msgs/srv/Automat
 # Stop automatic front collection
 ros2 service call /image_collection/toggle_front_collection auv_msgs/srv/AutomaticCapture "{data: false, time_interval: 0.0}"
 
-# This command securely copies all files from the remote Jetson's front camera data folder into your current local directory.
```
WilliamZhang2205, line 78: why was this comment removed?

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,258 @@ image_enhancement:
...
+    auto_focus: false                    # default true; when true, focus_absolute is ignored
+    focus_absolute: 7                    # [0..21]    only used when auto_focus is false; default 16
+
+    enable_object_detection: true
     enhanced_topic: /vision/down_cam/image_enhanced
     detection_topic: /vision/down_cam/detections
+    model_detection_threshold: 0.40
     # relative to src directory
-    model_relative_path: models/downcam.pt
+    model_relative_path: models/local_rfdetr_onnx_test # inference-models package directory (TODO: create this package)
```
WilliamZhang2205, line 175: why are we making a package for the models?

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,258 @@ image_enhancement:
...
+    detection_frame_id: zed_left_camera_frame # 3D detection frame used inside VisionDetectionFrame
+    image_frame_id: zed_left_camera_frame_optical # optical frame for image-like topics such as CameraInfo and debug feeds
+    auv_to_camera_center_xyz: [0.148, 0.0, 0.0]
+    auv_to_camera_center_rpy: [0.0, 0.0, 0.0]
+    camera_center_to_detection_xyz: [0.0, 0.06, 0.0]
+    camera_center_to_detection_rpy: [0.0, 0.0, 0.0]
+
+    # Detection-time filters / heuristics
+    enable_gate_top_crop: true # Removes bottom X% of bounding box for gate to ignore legs/pool floor noise
+    gate_top_crop_ratio: 0.50 # The ratio (0.0 to 1.0) of the gate bounding box to keep from the top edge
```
WilliamZhang2205, line 112: Why was this moved from game constraints to its own section? Why not group them in one section and then go through each object's filters like gate, octagon, pool limits

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,258 @@ image_enhancement:
...
+    model_relative_path: models/local_rfdetr_onnx_test # inference-models package directory (TODO: create this package)
     class_names:
       - "gate"
-      - "octagon_table"
-      - "octagon_top"
-      - "path_marker"
-      - "sawfish"
-      - "shark"
+      - "lane_marker"
+      - "red_pipe"
```
WilliamZhang2205, line 182: Mention that the class names here have to be manually maintained when changing models to have new classes. From what I see right now, it is not mentioned in the readme. If I recall properly, having a mismatch here can crash the vision node.

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
```diff
@@ -0,0 +1,310 @@
...
+        actual_h = int(cap.get(cv2.CAP_PROP_FRAME_HEIGHT))
+        actual_fps = cap.get(cv2.CAP_PROP_FPS)
+        self.node.get_logger().info(
+            f"Camera opened: {actual_w}x{actual_h} @ {actual_fps:.1f} fps"
+        )
+        return cap
+
+    def _apply_camera_controls(self):
+        """Apply V4L2 image controls from ROS parameters to the open camera."""
+        import subprocess
```
WilliamZhang2205: import at the top?

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
```diff
@@ -0,0 +1,310 @@
...
+            self._frame_counter += 1
+            frame_stamp = self.node.get_clock().now()
+
+            # ── Dataset collection ────────────────────────────
+            if self._collecting:
+                now = time.time()
+                if now - self._last_collection_time >= self.collection_interval:
+                    self._last_collection_time = now
+                    timestamp = datetime.now().strftime('%Y%m%d_%H%M%S_%f')
+                    filepath = os.path.join(self.collection_dir, f'down_{timestamp}.jpg')
```
WilliamZhang2205: Make png format/ test if changing to png can impact training performance or slow down image capture

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
```diff
@@ -0,0 +1,310 @@
...
+        self.node.get_logger().info(f"Applied camera controls: {', '.join(applied)}")
+
+    # ──────────────────────────────────────────────────────────────
+    # Main grab loop (runs on dedicated thread)
+    # ──────────────────────────────────────────────────────────────
+
+    def _grab_loop(self):
+        """Tight grab loop — runs on its own thread. read() blocks at camera fps."""
+        while rclpy.ok():
+            ret, img = self.cap.read()
```
WilliamZhang2205, line 286: As a note for the future, we could add an option to either grab the raw data or use the one from the down cam image enhancement. NO NEED FOR NOW. I think this is a good task for new member in the future since it is very "optimizing" our setup and currently image enhancement shouldn't be a priority since the RF-DETR model seems pretty good already.

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
```diff
@@ -0,0 +1,923 @@
...
+        init_params.camera_disable_self_calib = not self.zed_self_calib
+        init_params.camera_image_flip = self._enum_from_name(
+            sl.FLIP_MODE,
+            self.zed_camera_flip_mode_name,
+            'zed_camera_flip_mode',
+        )
+        init_params.enable_right_side_measure = self.zed_enable_right_side_measure
+        init_params.coordinate_units = sl.UNIT.METER
+        init_params.coordinate_system = sl.COORDINATE_SYSTEM.RIGHT_HANDED_Z_UP_X_FWD
+        init_params.depth_mode = sl.DEPTH_MODE.NEURAL
```
WilliamZhang2205: I think we can make the depth mode into a parameter. Don't know if we ever tried DEPTH_MODE.NEURAL_PLUS from https://www.stereolabs.com/docs/depth-sensing/depth-modes but I think it could be worth a shot to test it (computation usage vs accuracy) if not. Being able to switch between depth modes might be useful.

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
```diff
@@ -0,0 +1,923 @@
...
+        self.enable_object_detection = self.node.get_parameter('enable_object_detection').get_parameter_value().bool_value
+
+        self.conf_threshold = self.node.get_parameter('model_detection_threshold').get_parameter_value().double_value
+        self.depth_confidence_threshold = (
+            self.node.get_parameter('depth_confidence_threshold').get_parameter_value().integer_value
+        )
+        self.zed_depth_maximum_distance = (
+            self.node.get_parameter('zed_depth_maximum_distance').get_parameter_value().double_value
+        )
+        self.zed_depth_minimum_distance = (
```
WilliamZhang2205, line 222: add logic here to cap it to the ZED's maximums and minimums and make a log to say so. Even though the current params are indeed the minimums and maximums, this could help a new member testing the depth in the future to let them know what the limits are if they forget or if we ever manually change the current params.

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
```diff
@@ -0,0 +1,923 @@
...
+            self.node.get_parameter('zed_depth_minimum_distance').get_parameter_value().double_value
+        )
+        self.zed_positional_tracking_depth_min_range = (
+            self.node.get_parameter('zed_positional_tracking_depth_min_range').get_parameter_value().double_value
```
WilliamZhang2205, line 230: same here for adding a cap.

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
```diff
@@ -0,0 +1,923 @@
...
+        self.runtime_params.measure3D_reference_frame = sl.REFERENCE_FRAME.CAMERA
+        self.runtime_params.confidence_threshold = self.depth_confidence_threshold
+        self.runtime_params.texture_confidence_threshold  = 100
+
+        if not sim:
+            # Configure streaming
+            stream_params = sl.StreamingParameters()
+            stream_params.codec = sl.STREAMING_CODEC.H264
+            stream_params.port = stream_port
+            stream_params.bitrate = 8000 # kbps
```
WilliamZhang2205: curious but can this go higher?

WilliamZhang2205 (changes requested):
This is only half the review since PR is long. I will continue tmr and Thursday.

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: This is nor a ros2 package, it's a inference-models package

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Because it's used in front cam detection only now

Docker/jetson/down_cam.rules
kaidalisohaib: Yes it works

ros2_ws/src/sensors/src/State_aggregator.cpp
kaidalisohaib: For foxglove and general purpose if we need the transformation from pool_link to auv_pose

ros2_ws/src/sensors/params/sensors_frames.yaml
kaidalisohaib: For foxglove and general purpose if we need the transformation from pool_link to auv_pose

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
kaidalisohaib: We can try for sure

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
kaidalisohaib: Sounds good, good for future

Docker/jetson/requirements.txt
kaidalisohaib: Actually, we figured out that we should not pin them, because the jetson ai lab index url sometimes updates a package and removes the previous version

ros2_ws/src/sensors/src/State_aggregator.cpp
avyan-k: In that case, mention this here as a comment and in the README like under the table in [published topics](https://github.com/mcgill-robotics/AUV-2026/tree/dev/monolithic_vision_node/ros2_ws/src/sensors#published-topics), just so it's clear what this option does and how it is a easier ROS way to transform a pose from global frame to AUV frame

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,266 @@ image_enhancement:
 
 object_detection:
   front_cam:
+    # Topics / model
+    enable_object_detection: true
+    has_zed_sdk: true
     enhanced_topic: /vision/front_cam/image_enhanced # input topic if enhance_images is true, otherwise will subscribe to raw image topic
```
avyan-k: I know we are keeping it in case we will need but just for clarity it would good to mention when exactly this is used (i.e. for front cam it's:
```
enhanced_topic: ... # this will be the input topic if sim=true and no zed sdk AND general.enhance_image=true, otherwise see camera.front_cam_topic for input topic 
```
something like this maybe a little less wordy, just so if anyone decides to look into the image enhancement ROS nodes they aren't confused by this comment

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,266 @@ image_enhancement:
...
+    publish_annotated_image: true # publishes the front camera feed topic at detection_frame_topic + /annotated, toggled raw/annotated at runtime via service
+    publish_annotated_every_n_frames: 1 # publish the front camera feed every N frames (1 = every frame)
+    publish_depth_image: false # publish raw 32FC1 depth map image
+    publish_depth_compressed: false # publish depth map as compressed 16-bit PNG
+    publish_depth_every_n_frames: 5 # publish depth image every N frames
+
+    # Dataset collection
+    collection_dir: front_cam_collection # directory for direct-to-disk RGB image saving (toggled via service)
+    depth_collection_dir: depth_collection # directory for direct-to-disk 16-bit millimeter depth saving
+    collect_depth_image: false # flag to enable/disable depth saving
```
avyan-k: just to confirm have you guys tested image collection for front cam, depth down cam?  Did running all three (or two or anything you have tested) significantly impact the CPU)?

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,266 @@ image_enhancement:
...
+    white_balance_temperature: 4500      # [2800..6500] step 1; only used when auto_white_balance is false
+
+    # Exposure & focus
+    auto_exposure: 1                     # 1=Manual 3=Aperture Priority (auto); default 3
+    exposure_time_absolute: 40           # [4..5000]  only used when auto_exposure is 1 (Manual); default 625
+    auto_focus: false                    # default true; when true, focus_absolute is ignored
+    focus_absolute: 7                    # [0..21]    only used when auto_focus is false; default 16
+
+    enable_object_detection: true
     enhanced_topic: /vision/down_cam/image_enhanced
```
avyan-k: same comment here as with front enhanced

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,266 @@ image_enhancement:
...
+      - "sos"
+      - "ambulance"
+      - "firetruck"
+      - "blood"
+      - "fire"
+      - "bin_structure"
     queue_size: 10
-    publish_annotated_image: true # whether to publish an annotated image with bounding boxes drawn (for visualization/debugging). will publish to detection_topic + /annotated
-    log_level: warn
+    model_detection_threshold: 0.40
```
avyan-k, line 49: 0.4 seems a little insane to me were there no false positives? Have you tried visualizing the ROC curve over the synth dataset and/or last year comp dataset? It should give an even better idea of where to threshold

ros2_ws/src/vision/config/vision_pipeline.yaml
```diff
@@ -16,73 +16,266 @@ image_enhancement:
...
+    # Exposure & focus
+    auto_exposure: 1                     # 1=Manual 3=Aperture Priority (auto); default 3
+    exposure_time_absolute: 40           # [4..5000]  only used when auto_exposure is 1 (Manual); default 625
+    auto_focus: false                    # default true; when true, focus_absolute is ignored
+    focus_absolute: 7                    # [0..21]    only used when auto_focus is false; default 16
+
+    enable_object_detection: true
     enhanced_topic: /vision/down_cam/image_enhanced
     detection_topic: /vision/down_cam/detections
+    model_detection_threshold: 0.40
```
avyan-k, line 173: same comment as with front cam model this is isn't strictly necessary for the merge but please look into this if you have not already

ros2_ws/src/vision/config/vision_pipeline.yaml
avyan-k, line 1: NOT A CHANGE REQUEST DON'T CRASH OUT SOHAIB
The current state of adding parameters to the vision pipeline is much more tedious than it needs to be. To add a parameter for a given camera, you need to add it to config yaml, then the vision pipeline launch then the specific node it is relevant for. The typically ROS way to do this is pass the params directly to the relevant node, as mentioned in planner the yaml safe load trick should mostly be done if parameters need to be kept consistent between multiple nodes (like object detection output topic  == object map input topic). This is honestly a mistake from when I originally wrote the vision pipeline because I was not as familiar with standard ROS practices and so believed each node should have it's own config. I could not touch vision pipeline until this branch is reviewed and merged but I will probably make a PR to simplify all this because I expect there to more params to be added to the vision nodes as the needs of planner become clear

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
```diff
@@ -418,65 +521,153 @@ void ObjectTracker::create_new_tracks(
...
-                // Apply a very high confidence measurement update to pull the gate to the fish midpoint
-                Eigen::Matrix3d tiny_R = Eigen::Matrix3d::Identity() * 0.001;
-                gate_track->kf.update(midpoint, tiny_R);
-                // Update p_gate for orientation calculation
-                p_gate = gate_track->kf.state();
-            }
+        Eigen::Vector3d midpoint = (p_search_rescue + p_survey_repair) / 2.0;
+
+        if ((p_gate - midpoint).norm() < kRefinementPlausibilityRadiusMeters) {
+            Eigen::Matrix3d tiny_R = Eigen::Matrix3d::Identity() * 0.001;
```
avyan-k: similar to epsilon = 1e-6 you could define a const expr for this tiny covariance matrix as well

ros2_ws/src/vision/vision/object_map.cpp
```diff
@@ -1,496 +1,537 @@
...
+        filtered_covariances.reserve(msg->detections.size());
+        filtered_classes.reserve(msg->detections.size());
+        filtered_orientations.reserve(msg->detections.size());
+        filtered_confidences.reserve(msg->detections.size());
+
+        for (const auto& detection : msg->detections) {
+            const std::string& label = detection.label;
+
+            Eigen::Vector3d pos_camera = eigen_from_point(detection.pose_camera.pose.position);
+            if (label == "red_pipe" || label == "white_pipe") {
```
avyan-k: I think as with every other game specific constraint this should be an option. I think it would be really cool to have a "before" and "after" which shows how the filtering cleans up everything (it would even be nice to have this in the report like raw image feed to detections to map or something like this

ros2_ws/src/vision/vision/object_map.cpp
```diff
@@ -1,496 +1,537 @@
...
+
+        Eigen::Matrix3d camera_to_world_rotation;
+        for (int r = 0; r < 3; ++r) {
+            for (int c = 0; c < 3; ++c) {
+                camera_to_world_rotation(r, c) = world_camera_basis[r][c];
+            }
+        }
+
+        Eigen::Vector3d observer_position = eigen_from_point(msg->auv_pose.pose.position);
+        bool has_observer_position = true;
```
avyan-k, line 265: I'm assuming this field (and where it is used thereafter in the orientation additions) is just in case the observer position becomes stale or invalid, but from what I see it always remains true, correct?

ros2_ws/src/vision/models/local_rfdetr_onnx_test/inference_config.json
```diff
@@ -0,0 +1,30 @@
...
+            "width": 512
+        },
+        "dynamic_spatial_size_supported": false,
+        "color_mode": "rgb",
+        "resize_mode": "letterbox",
+        "input_channels": 3,
+        "scaling_factor": 255.0,
+        "normalization": [
+            [
+                0.485,
```
avyan-k, line 15: As we discussed it would be nice to test different normalization parameters notably taking the avg and std of the data synth dataset and potential the pool dataset if we have the new props (though technically speaking props shouldn't impact these parameters so using our existing pool dataset should be fine)

ros2_ws/src/vision/models/template_rfdetr/inference_config.json
```diff
@@ -0,0 +1,30 @@
...
+            "width": 512
+        },
+        "dynamic_spatial_size_supported": false,
+        "color_mode": "rgb",
+        "resize_mode": "letterbox",
+        "input_channels": 3,
+        "scaling_factor": 255.0,
+        "normalization": [
+            [
+                0.485,
```
avyan-k, line 15: same comment as with YOLO template

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
```diff
@@ -1,30 +1,103 @@
 #include <object_tracker.hpp>
 #include <hungarian.hpp>
 
+#include <cmath>
+
+namespace {
+
+constexpr double kRefinementPlausibilityRadiusMeters = 5.0;
```
avyan-k: make this into a param in at least the constructor of tracker and ideally in the object map node so we can play around with it if needed

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
```diff
@@ -1,30 +1,103 @@
...
+
+Eigen::Vector2d xy(const Eigen::Vector3d& position) {
+    return position.head<2>();
+}
+
+double xy_distance(const Eigen::Vector3d& lhs, const Eigen::Vector3d& rhs) {
+    return (xy(lhs) - xy(rhs)).norm();
+}
+
+bool compute_facing_normal(
```
avyan-k, line 24: Very clean implementation, I couldn't mention this in the original object map PR since I worked on it but I love this. I'm sure you've tested this works already but I can confirm this should because this is essentially what we had to do for collision detection in COMP 521. This isn't a change request but it's so cool seeing this here (I can't link my unity implementation bcs it's private as it contains assignment instructions but for example:
```C#
public static Vector2 NormalClosestToPoint(Vector2 Point, Vector2 LineA, Vector2 LineB)
  {
      // normal = (-dy,dx)
      Vector2 normal = new(LineA.y - LineB.y, LineB.x - LineA.x);
      if (Vector2.Dot(normal, Point - LineA) == 0) throw new InvalidGeometryException($"Point {Point} is colinear to line through {LineA} and {LineB}. Unable to determine normal.");
      // if vector pointing towards points is similar to normal, than that is closer, otherwise opposite
      return Vector2.Dot(normal, Point - LineA) > 0 ? normal : -normal;

  }
```
is almost word for word what is written here. It might be a bit obvious but still very cool!)

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
```diff
@@ -1,30 +1,103 @@
...
+    const Eigen::Vector2d& axis_xy,
+    const Eigen::Vector2d& object_xy,
+    const Eigen::Vector2d& observer_xy,
+    Eigen::Vector2d& chosen_normal)
+{
+    if (axis_xy.norm() < 1e-6) {
```
avyan-k: magic number (along with the 1e-6 below). because this is just an measurement error epsilon  a `constexpr double` makes sense here, a global indication when you consider two vectors to be the same

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
avyan-k: Remove this if you believe matching track like this is appropriate (including the debug print below)

avyan-k: Not sure what the consensus ended up being but frame based age is overall better imo so just remove this comment to avoid confusion

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
```diff
@@ -388,11 +467,35 @@ void ObjectTracker::create_new_tracks(
         for (const auto& existing_track : tracks) {
             Eigen::Vector3d existing_pos = existing_track.kf.state();
             double dist = (new_pos - existing_pos).norm();
-            if (dist < min_new_track_distance) {
+            if (existing_track.label == label && dist < min_new_track_distance) {
```
avyan-k: shouldn't we check if it's the same label before doing the difference computations (like `continue` on differing labels) ? I know computationally it's small optimization but might as well and it's more readable to skip first

ros2_ws/src/vision/vision/object_map.cpp
```diff
@@ -1,496 +1,537 @@
...
+        lane_y_max = this->declare_parameter<double>("lane_y_max", 100.0);
+
+        if (enable_lane_boundary) {
+            RCLCPP_INFO(
+                this->get_logger(),
+                "Lane boundary enabled: x=[%.1f, %.1f] y=[%.1f, %.1f]",
+                lane_x_min, lane_x_max, lane_y_min, lane_y_max);
+        }
+
+        const int enabled_table_octagon_modes =
```
avyan-k: Instead of having three toggles why not just have 4  refinement modes, either 
table is true xy
octagon is true xy
midpoint of table and octagon is true xy
no refinement
more readable and prevents accidental toggles

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
-## Launch File Usage
-In most cases, the [vision pipeline](#vision-pipeline) launch file will be the main entry point for running vision nodes. All other launch files are intended for testing and development of individual nodes, and are run as intermediate steps in the vision pipeline. It expects models to already be trained (see [Steps to train a model](./model_pipeline/README.md)) and placed in the `ros2_ws/vision/models` directory.
-### VIsion Pipeline
-Launches the entire vision pipeline, image enhancement -> object detection -> object mapping.
-
-```bash
-ros2 launch vision vision_pipeline.launch.py
+This package contains the front-camera detection pipeline, the persistent 3D object map, optional image enhancement nodes, and the model-training/export helpers used to build the deployed detectors.
+
+The current runtime center of gravity is:
```
avyan-k: what? keep language simple and descriptive

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+```mermaid
+flowchart LR
+    A["Front camera / ZED<br/>RGB + depth source"] --> B["front_cam_object_detection<br/>infer + ZED 3D + frame sync"]
+    S["sensors / state_aggregator<br/>official pose + TF"] -->|"state/pose"| B
+    B -->|"VisionDetectionFrame"| C["object_map<br/>camera->world + tracking"]
+    B -->|"front camera feed + CameraInfo"| F["Foxglove / debug<br/>visualization"]
+    B -->|"optional /vision/vio_pose"| E["debug / future fusion<br/>vision pose measurement"]
+    T["Static TF<br/>auv_link -> zed_camera_center -> zed_left_camera_frame -> optical"] --> C
+    S -->|"official TF: pool_link -> auv_link"| C
+    C -->|"VisionObjectArray"| D["/vision/object_map<br/>persistent world map"]
```
avyan-k, line 52: Since this is the main output of the vision package it would be good to highlight it. This can be done in mermaid flowcharts with class description
```
classDef spotlight fill:#ffff,stroke:#FF0000,stroke-width:4px,color:#1f1f1f,font-weight:bold;
class D spotlight;
```
add this to the end of the code block and adjust colors as necessary

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+- `Front camera / ZED`: Provides the raw RGB images and ZED depth data used to build 3D detections.
+- `sensors / state_aggregator`: Publishes the official aggregated AUV pose and the official `pool_link -> auv_link` TF.
+- `front_cam_object_detection`: Runs the detector, asks ZED for 3D object positions, and packages each frame with one synchronized AUV pose.
+- `Static TF`: Describes the fixed mounting relationship between the AUV body frame and the front camera frames.
+- `object_map`: Converts camera-frame detections into world-frame objects and tracks them over time.
+- `/vision/object_map`: Exposes the persistent world-frame map as `VisionObjectArray`.
+- `Foxglove / debug`: Visualizes the front camera feed, camera calibration, and mapped detections.
+- `debug / future fusion`: Carries the optional vision-only VIO pose output for debugging or later estimator fusion.
+
+### Front camera path
```
avyan-k: path?

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+- [VisionDetection.msg](../auv_msgs/msg/VisionDetection.msg)
+- [VisionDetectionFrame.msg](../auv_msgs/msg/VisionDetectionFrame.msg)
+
+`VisionDetectionFrame` is the input to `object_map`. It carries:
+- `header.stamp`: frame timestamp
+- `header.frame_id`: 3D detection frame, currently `zed_left_camera_frame`
+- `auv_pose`: AUV pose snapshot for that frame, typically from `state/pose`
+- `detections`: camera-frame 3D detections plus the original model 2D bbox
+
+### Object map path
```
avyan-k: path...?

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+The object map node:
+- subscribes to `VisionDetectionFrame`
+- resolves the TF chain from `auv_link` to `zed_left_camera_frame`
+- transforms detections from camera frame to world frame in C++
+- tracks objects over time with a Kalman + Hungarian pipeline
+- publishes the persistent map as:
+  - [VisionObject.msg](../auv_msgs/msg/VisionObject.msg)
+  - [VisionObjectArray.msg](../auv_msgs/msg/VisionObjectArray.msg)
+
+### How the front camera node works
```
avyan-k: why not make this a subheader of front cam node

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+3. run detector inference on the RGB image
+4. optionally filter detections near image borders
+5. optionally crop the lower half of gate boxes before depth ingestion
+6. ingest the 2D boxes into the ZED SDK
+7. retrieve 3D detections in `zed_left_camera_frame`
+8. publish one `VisionDetectionFrame`
+
+This keeps the expensive 2D-to-3D step close to the ZED SDK while still leaving world-frame conversion to `object_map`.
+
+### How object_map works
```
avyan-k: why not make this a subheader of object map node

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
-Select camera [f]ront or [d]own: f
-
-Action: c  (capture image)
-Action: v  (view latest - Jetson only)
-Action: b  (back to main menu)
-```
-
-##### Automatic Capture Workflow
-
+ros2 launch vision vision_pipeline.launch.py sim:=true
```
avyan-k: Please make a table for all launch parameters since that will be the primary way to interact with the package. For the same reason usage/installation is typically at the top of a README not buried a third of the way in..

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+- static TF publishers for the front camera
+- object map
 
-Images are saved with timestamps:
-- `data_front_cam/image_YYYYMMDD_HHMMSS_microseconds.jpg`
-- `data_down_cam/image_YYYYMMDD_HHMMSS_microseconds.jpg`
+The down-camera nodes still exist, but they are not part of the default runtime pipeline in `vision_pipeline.launch.py`.
 
-#### Parameters
+## Important Topics
```
avyan-k: what does important topics means here? Sensors and controls have input (subscription) and output (publishing) topics and that much more clearer, just include everything in a table (or bullet points ig)

ros2_ws/src/vision/README.md
```diff
@@ -1,178 +1,341 @@
...
+- `/image_collection/toggle_manual_front_collection`
+  - compatibility alias for the same `AutomaticCapture` behavior
 
-This will launch the image enhancement nodes for both cameras with default parameters. Arguments can be set in the launch file or via command line:
+Example request:
 
-```bash
-ros2 launch vision image_enhancement.launch.xml PARAM_NAME:=VALUE
+```json
+{ "data": true, "time_interval": 1.0 }
```
avyan-k: include the CLI service call command

avyan-k (changes requested):
Reviewed vision package, looks great. The comments on the README are not strictly necessary for merge as I'll probably do a PR for the vision pipeline simplification and will probably include a documentation revamp. 

Also to simplify the documentation work for you guys I will add a commit after this review bringing back code comments that were lost from the previous object map implementation (specifically in object tracker a lot of nitpicky stuff so I found it simpler to add it myself as I was reviewing)

ros2_ws/src/sensors/src/State_aggregator.cpp
kaidalisohaib: Done

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Done

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Yes, I tried all 3 of them at the same time and no significant impact on the CPU was seen, probably because it's saving every 1sec at maximum and the images are relatively small

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map.cpp
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map.cpp
kaidalisohaib: I added a comment to it, it basically makes the code more readable by explicitly checking if it does have an auv pose to perform the calculation needed rather than assuming the observer is always at a valid coordinate on the tracker side and future proofing in case we want to test object map with a camera which does not have a position.

ros2_ws/src/vision/models/local_rfdetr_onnx_test/inference_config.json
kaidalisohaib: I looked it up and my instinct was right about keeping the values the same for our situation. Since RF-DETR uses a DINOv2 backbone that was pretrained with ImageNet normalization, the model's feature representations are literally conditioned on these specific mean/std values. Changing them would shift the input distribution away from what the weights expect and would actually hurt performance. The threshold for switching to custom dataset stats is for extreme domain shifts like thermal or medical imaging, not underwater RGB which is still natural images just with a color cast. So we should keep the ImageNet defaults as is

ros2_ws/src/vision/models/template_rfdetr/inference_config.json
kaidalisohaib: YOLO only uses scaling_factor 255.0 (dividing pixel values by 255 to get the 0-1 range) and that's it, no mean/std normalization

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
kaidalisohaib: Done, removed the `TODO:`

kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map/object_tracker.cpp
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_map.cpp
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/README.md
kaidalisohaib: Done

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Sounds good, basically have one config file per node and pass it to the launch node to make it easier to read and add parameters :+1:

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
kaidalisohaib: They have recommended bit rates here:

https://www.stereolabs.com/docs/video/streaming

We can experiment with more, but my guess is that since we're using a low resolution we won't need more.

Docker/jetson/requirements.txt
kaidalisohaib: Ideally, we would have a more stable index url that also keeps previous versions to not run into a cycle of having to update the version each time the server updates it.

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
kaidalisohaib: Done

ros2_ws/src/vision/vision/object_detection/down_cam_object_detection_node.py
kaidalisohaib: From my experimentation, every time it saves an image with `.png` the latency of the current frame goes up by ~50ms, which I think is worth it since it's happening every 30 frames if we collect images every 1sec and is not that big of a hit on performance. I changed it to `.png`

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: The current model is around 0.90 precision / 0.84 recall on the synthetic validation set. We’ll still need to tune this threshold empirically, but initial testing on last year’s competition images at a either 0.4 or 0.5 threshold produced relatively few false positives while maintaining good detections without fine tuning the model

ros2_ws/src/vision/config/vision_pipeline.yaml
kaidalisohaib: Answer in previous comment

kaidalisohaib:
image

Please guys no more comments :pray:

ArthurGuihaire (dismissed):

avyan-k (dismissed):
LGTM hopefully passes CI and after William finishes review we can merge 🔥

JPGC04:
> image
> 
> Please guys no more comments :pray: 

Okay bro

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
```diff
@@ -0,0 +1,1033 @@
...
+
+                    bbox_xyxy = model_bboxes_by_unique_id.get(obj.unique_object_id)
+                    if bbox_xyxy is not None:
+                        (
+                            detection.bbox_center_x,
+                            detection.bbox_center_y,
+                            detection.bbox_size_x,
+                            detection.bbox_size_y,
+                        ) = bbox_from_xyxy(*bbox_xyxy)
+                    else:
```
WilliamZhang2205, line 997: In which case would an id object from the zed not be one of the ones already in the map? I looked up the docs and they do not seem to mention too many details on that. I asked GPT and it says it could be because ZED has its own tracker that could persist between two frames. Wondering if that's true and you guys did see it.

If that is the case,  generating new unique IDs per detection frame would be a potential source of error since maybe that ID already exists in the ZED tracking software between two frames. I could be wrong and assuming the way it works to be wrong. If you guys could clear it up for me itd be great!

ros2_ws/src/vision/vision/object_map.cpp
```diff
@@ -1,496 +1,545 @@
...
+                false,
+                0.0,
+                octagon_confidence,
+                octagon_age));
+        }
+
+        object_map_publisher->publish(object_map_msg);
+        rclcpp::Time pipeline_end_time = this->now();
+        rclcpp::Duration time_diff = pipeline_end_time - frame_collection_time;
+        RCLCPP_INFO(
```
WilliamZhang2205: debug only?

ros2_ws/src/vision/vision/object_map.cpp
```diff
@@ -1,496 +1,545 @@
...
+            this->get_logger(),
+            "Object map pipeline latency: %.9f seconds",
+            time_diff.seconds());
+    }
+
+    std::map<std::string, Track> persistent_objects;
```
WilliamZhang2205, line 533: In a future polished version, make a header and put these there.

ros2_ws/src/vision/models/template_rfdetr/inference_config.json
WilliamZhang2205: explain origin and meaning of these files in the readme?

WilliamZhang2205 (changes requested):

ros2_ws/src/vision/vision/object_detection/front_cam_object_detection_node.py
kaidalisohaib: Since we have the tracking disabled it should be a 1:1 match. The else block is just defensive, if the ZED SDK clips a box at the image edge or the ID mapping fails during its internal rectification, we can still use the ZED corners so we don't publish empty boxes. It's basically a safety net for edge cases

ros2_ws/src/vision/vision/object_map.cpp
kaidalisohaib: Sounds good :+1:

ros2_ws/src/vision/vision/object_map.cpp
kaidalisohaib: Done

ArthurGuihaire (dismissed):

ros2_ws/src/vision/models/template_rfdetr/inference_config.json
kaidalisohaib: DONEEEEE

ArthurGuihaire (approved):
lgtm

avyan-k (approved):

WilliamZhang2205 (approved):
👍👍👍

PR #106, opened 2026-05-13
Simplify vision launch files and parameter management

avyan-k:
Update on #103

**Minor file reorganization:**
* Moved `hungarian.cpp` and `kalman.cpp` from `utils/` to `external/` in the `object_map_lib` library, reflecting a clearer separation of internal and external dependencies.

* Extracted object map headers into separate [object_map_node.hpp](https://github.com/mcgill-robotics/AUV-2026/blob/dev/simplify-vision-pipeline/ros2_ws/src/vision/include/vision/object_map_node.hpp) file and `main` function into it's own separate entrypoint at [object_map.cpp](https://github.com/mcgill-robotics/AUV-2026/blob/dev/simplify-vision-pipeline/ros2_ws/src/vision/vision/object_map.cpp)

**Node and header additions:**
* Refactor `vision_pipeline.yaml` to use hierarchical parameters. Since this will be the main interface for the vision pipeline, using a hierarchy structure improves readability and refactoring
* Refactored `vision_pipeline.launch.py` to use the new parameter and topic structure, updating default values and topic resolution to align with the updated configuration files. Also renamed enhancement node names and log level handling for consistency. [[1]](diffhunk://#diff-8f3263f5c208ea94a1f260b0a80ad4e0fd56fccd23454982f9758055a47271a5L37-R67) [[2]](diffhunk://#diff-8f3263f5c208ea94a1f260b0a80ad4e0fd56fccd23454982f9758055a47271a5L79-R123)

**Object tracker enhancements:**
* Added new boolean parameters to the `ObjectTracker` class for enabling/disabling large structure separation and large structure/pipe separation, providing finer control over tracking behavior. [[1]](diffhunk://#diff-7a76e42db50e9e3d553c949b03960185ccb078b61fda2d8ee7c4dc64bbecb001R53-R55) [[2]](diffhunk://#diff-7a76e42db50e9e3d553c949b03960185ccb078b61fda2d8ee7c4dc64bbecb001R140-R142)


kaidalisohaib (dismissed):
While the hierarchical changes in the config look good visually and improve readability, in practice, it will make it harder to find and change parameters quickly. The same applies to the corresponding README updates that reflect these new structures. Additionally, it looks like many files overwrote their git history, though that's not a major issue in the grand scheme of things since we know who did what in between ourselves

While I don't fully agree with the hierarchical parameter approach, I'm approving this so we can get it merged and move on to more impactful changes

avyan-k:
> While the hierarchical changes in the config look good visually and improve readability, in practice, it will make it harder to find and change parameters quickly. The same applies to the corresponding README updates that reflect these new structures. Additionally, it looks like many files overwrote their git history, though that's not a major issue in the grand scheme of things since we know who did what in between ourselves
> 
> While I don't fully agree with the hierarchical parameter approach, I'm approving this so we can get it merged and move on to more impactful changes

Alright, reverted most of the hierarchical parameters and tested to make sure it works in sim. I still kept a kept a few "categories" (e.g. some of the smaller filters). I will retest this on the jetson to reconfirm especially the down cam changes

As for the git blame, unfortunately git cannot determine a file rename if (I believe) over 50% of the file was change during the rename. However most editors (e.g. VSCode) are still able to smartly determine the blame and thus the history so it should be okay

kaidalisohaib (dismissed):
For the git history I was also referring to the copy pasting in object_map.cpp. But honestly, it's completely fine, let's just get this merged as quickly as possible, this PR was optional anyway, and there are more important features I want to add to the object map soon
Thanks for reverting the parameters

kaidalisohaib (approved):
lgtm

ArthurGuihaire (approved):
lgtm, not necessary but does a nice job cleaning up the configs

PR #108, opened 2026-05-17
Controls V2

### LINKED ISSUE #93: Address non-linearity in thruster output for large wrenches
Address non-linearity in thruster output for wrenches with high magnitude. 

Example: 
We have an arbitrary wrench with non-zero elements on all 6 degrees of freedom. How do we solve the for the array of optimal thruster output without "distorting" the desired wrench. Say we have a wrench of [200 N,200 N,200 N, 50 N.m, 0, 60 N.m]. We clearly want to move diagonally, dive, and yaw. With our maximum thruster output of 40N, we cannot possibly satisfy the desired wrench. 

Solving the system of equations with the allocation matrix yields a result of .How can we account for such limitations of the thrusters without "distorting" the desired wrench?

### LINKED ISSUE #98: Update Allocation Matrix based on New Thruster Mounts
The AUV now has new thruster mounts, which altered the thrusters' position with respect to the AUV's center of mass. 

Goal Objectives:

- Update the Allocation Matrix to reflect the thrusters' new mounting positions

- [Optional] Clean up and organize all propulsion parameters into a single .yaml file for better readability and maintainability.

### LINKED ISSUE #101: Update IMU processing to accomodate new IMU mounting orientation
The IMU is now mounted upside-down in the AUV. The imu processor needs to be updated accordingly

nadimasmar:
This PR is for a improved version of the controls package. Main improvements are:

- Reworked Propulsion package to solve for thruster forces while respecting thruster limits. Our current implementation assumes the thrusters have unlimited thrust capacity, which is problematic when doing maneuvers that utilize the thrusters to their limits. Details on the mathematical implementation can be found in the docs of the propulsion package. **Propulsion now has an added dependency: [OSQP Python](https://github.com/osqp/osqp)**

- Controls now has two seperate attitude controllers. One for holding an attitude, and another dedicated for maneuvers and large errors. This allows us to tune each controller independantly so each can achieve their resepctive functions. 

- Added a `trajectory_planner` node. This node manages and provides an interface to command flips. It uses a minimum-jerk polynomial to generate a smooth flip trajectory for the AUV to follow.


ros2_ws/src/propulsion/params/propulsion.yaml
```diff
@@ -0,0 +1,39 @@
...
+    max_thrust_t2: 40.0
+    max_thrust_t3: 40.0
+    max_thrust_t4: 40.0
+    max_thrust_t5: 40.0
+    max_thrust_t6: 40.0
+    max_thrust_t7: 40.0
+    max_thrust_t8: 40.0
+
+    # Minimum thrust values that can be commanded for each thruster (in Newtons)
+    min_thrust_t1: -20.0
```
JPGC04, line 14: Min and max thrusts should be the same values but in inverse because if you are moving forward for example and each thruster needs to apply 40 Newtons in its own axis, you don't want 2 thruster applying 40 Newtons and the other 2 applying -20 Newtons. It will lag if I am not mistaken.

ros2_ws/src/propulsion/params/propulsion.yaml
```diff
@@ -0,0 +1,39 @@
...
+    min_thrust_t2: -20.0
+    min_thrust_t3: -20.0
+    min_thrust_t4: -20.0
+    min_thrust_t5: -20.0
+    min_thrust_t6: -20.0
+    min_thrust_t7: -20.0
+    min_thrust_t8: -20.0
+
+    # Thruster PWM limits in microseconds
+    thruster_PWM_lower_limit: 1100
```
JPGC04: PWM upper and lower limits need to match those we were using before which are 1228 to 1768. I don't remember how the values were chosen, but they were a clamp on the safe amount of current allowed per thruster.

ros2_ws/src/propulsion/params/propulsion.yaml
```diff
@@ -0,0 +1,39 @@
...
+    thruster_PWM_upper_limit: 1900
+
+    # Safety
+    watchdog_timeout_sec: 2.0   # Reset thrusters if no wrench received for this long
+    arming_delay_sec: 1.0       # Seconds to wait before accepting wrench commands
```
JPGC04, line 29: I would like arming delay to always be 0, the armign delay is already accounted for in embedded. This is something we have been planing to remove for the past year but did not for the sake of legacy. But since it is this easy to remove now, please set it at 0.

ros2_ws/src/propulsion/propulsion/thrust_mapper.py
```diff
@@ -41,31 +42,67 @@ def __init__(self):
...
+        self.declare_parameter('min_thrust_t1', -40.0)
+        self.declare_parameter('min_thrust_t2', -40.0)
+        self.declare_parameter('min_thrust_t3', -40.0)
+        self.declare_parameter('min_thrust_t4', -40.0)
+        self.declare_parameter('min_thrust_t5', -40.0)
+        self.declare_parameter('min_thrust_t6', -40.0)
+        self.declare_parameter('min_thrust_t7', -40.0)
+        self.declare_parameter('min_thrust_t8', -40.0)
+
+        self.declare_parameter('thruster_PWM_lower_limit', 1100)
```
JPGC04, line 66: Please update those parameters as I indicated above

ros2_ws/src/propulsion/propulsion/thrust_mapper.py
```diff
@@ -148,53 +214,42 @@ def wrench_to_thrust(self, wrench_msg: Wrench):
...
         pwm_arr = np.clip(pwm_arr, self.thruster_lower_limit, self.thruster_upper_limit)
-        pwm_arr = pwm_arr.astype(np.uint16, copy=False) # Ensure uint16 for message compatibility
-        
+        pwm_arr = pwm_arr.astype(np.uint16, copy=False)
+
         pwm_msg = Int16MultiArray()
         pwm_msg.data = pwm_arr.tolist()
         self.pub_us.publish(pwm_msg)
 
     def re_arm(self):
```
JPGC04, line 248: Thank you for implementing this change. Sending 1500 is enough to arm. When we used to send 1540 then 1500 it was always a risk to someone who's fingers are close to the thrusters and then we arm and then the thurster spins near their finger.

ros2_ws/src/propulsion/params/propulsion.yaml
kaidalisohaib: Isn't because the thrusters are more capable in the forward thrust compared to backward?

ros2_ws/src/propulsion/params/propulsion.yaml
kaidalisohaib: The solver needs to have control over the lower end of the limits and the previous PWM limits made it such that the thrusters would've gotten saturated before the solver reached the actual limit wanted by the solver.

ros2_ws/src/propulsion/params/propulsion.yaml
JPGC04: Yes, but if we do need 40 per thruster, then there will be a lag between thrusters pushing forwards and thruster pushign backwards. If the thruster cannot exceed -20 in backward direction then the upward limit would have to be 20 too.

ros2_ws/src/propulsion/params/propulsion.yaml
JPGC04: There still has to be a clamp for elec purposes or Dougie will overheat. Upon discussion with Nadim we determined clamps to be at 1196 to 1804. Unfortunately operating with no clamps to the pwm is simply unsafe for operation as there is a safety margin + max current we allow for each thruster before its fuse pops.

ros2_ws/src/propulsion/params/propulsion.yaml
JPGC04: After further discussion with Sohaib, there timeout should remain at 1 to ensure the thrusters do recieve the arming signal. It can be modified in the future if we ever need to.

ros2_ws/src/controls/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,129 @@
...
+                                msg->x,
+                                msg->y,
+                                msg->z
+                        );
+                }
+        }
+
+        void TrajectoryPlanner::flip_command_callback(const auv_msgs::msg::FlipCommand::SharedPtr msg)
+        {
+                flip_in_progress_ = true;
```
avyan-k, line 46: In my opinion the flip command should be a service rather than a subscription to the topic. This would make it more consistent with ROS standards where services/actions are designed for commands. It would also allows rejection of commands such as if a flip is not feasible / advisable (this does not need to be implemented here, but it is good to have the setup for it) and if the axis provided is not invalid

ros2_ws/src/controls/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,140 @@
...
+                msg.angular_velocity.x = w_ref_v.x();
+                msg.angular_velocity.y = w_ref_v.y();
+                msg.angular_velocity.z = w_ref_v.z();
+                pub_reference_attitude_->publish(msg);
+        }
+
+        float TrajectoryPlanner::minimum_jerk(const float u) const
+        {
+
+                return 10 * std::pow(u, 3) - 15 * std::pow(u, 4) + 6 * std::pow(u, 5);
```
WilliamZhang2205, line 124: what are the origins of theses values? Reasons being: One for documentation and two if someone wants to change this later on

ros2_ws/src/controls/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        float TrajectoryPlanner::minimum_jerk_derivative(const float u) const
+        {
+                return 30 * std::pow(u, 2) - 60 * std::pow(u, 3) + 30 * std::pow(u, 4);
```
WilliamZhang2205, line 129: same as above

ros2_ws/src/auv_msgs/msg/FlipCommand.msg
```diff
@@ -0,0 +1,18 @@
+# FlipCommand.msg
+
+# Axis of rotation in the AUV body frame
+uint8 AXIS_X=0
+uint8 AXIS_Y=1
+uint8 AXIS_Z=2
```
WilliamZhang2205, line 6: Just as an aside idea, we could in the future change this message, or make a new one to be able to make flips along any desired vector. Of course this requires more work in trajectory planner, but itd be a cool feature to have

ros2_ws/src/auv_msgs/msg/AttitudeReference.msg
```diff
@@ -0,0 +1,7 @@
+# auv_msgs/msg/AttitudeReference.msg
+
```
WilliamZhang2205, line 2: Add a comment describing the reference oritentation to the orientation variable. A first time reader might assume it is the AUV frame, but a clear comment would be nice

ros2_ws/src/controls/include/controls/trajectory_planner.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+                void publish_attitude_reference(const quatd& q, const Vec3& w_ref_v) const;
+                float minimum_jerk(float u) const;
+                float minimum_jerk_derivative(float u) const;
+
+                quatd q_start_ = quatd::Identity();
+
+                bool flip_in_progress_ = false;
+
+                float planner_rate_hz_ = 10.0;
```
WilliamZhang2205: maybe make this a parameter from configs and nitpicky but could it be trajectory_planner_rate_hz? Id like to clear any confusion that could arise between this planner and the planner package which has its own frequency.

ros2_ws/src/controls/src/attitude_controller.cpp
```diff
@@ -72,30 +104,50 @@ namespace controls
...
-        q_error = sensors::math::canonicalizeShortest(q_error);
-        Vec3 error_vector = Vec3(q_error.x(), q_error.y(), q_error.z());
-        Vec3 feedback = P_e_ * error_vector - P_w_ * w_iv_;
+
+        Vec3 feedback = Vec3::Zero();
+
+        switch(mode)
+        {
+            case ControlMode::LARGE_ERROR:
+            {
```
WilliamZhang2205, line 178: add comment about the shortest quaternion being properly set in the compute control effort.

ros2_ws/src/controls/src/attitude_controller.cpp
avyan-k, line 1: Not sure if this has been done already, but it would be nice to update the attitude controls PDF in docs with the new split attitude controls (i.e. holding attitude vs maneuvers). The existing doc was very useful for explaining the math behind the attitude controller, so if it has been updated please upload it to this PR so anyone working on controls in the future will have an easier time parsing the logic of the code

ros2_ws/src/propulsion/propulsion/thrust_allocator.py
```diff
@@ -0,0 +1,175 @@
...
+
+        self.qp_solver.update(
+            l=np.concatenate([target_wrench, self.min_thrust]),
+            u=np.concatenate([target_wrench, self.max_thrust]),
+        )
+
+        results = self.qp_solver.solve()
+
+        if results.info.status_val != osqp.constant("OSQP_SOLVED"):
+            raise RuntimeError(f"QP solver failed with status: {results.info.status}")
```
avyan-k, line 154: This should probably not raise a runtime error but ideally throw a custom exception which can be caught by the ROS node for logging etc. Right now the caller ROS node catches all runtime exceptions and simply logs then which may be a problem since there may be other logic or code issues that lead to runtime exception which **should** crash the node.

ros2_ws/src/controls/src/superimposer.cpp
```diff
@@ -111,38 +139,68 @@ namespace controls
...
+       if (!enabled_) {
+           return;
+       }
+
        wrench_msg combined_effort;
 
        // Transform depth effort from pool frame to body frame. 
        Vec3 depth_force_pool(depth_effort_.force.x, depth_effort_.force.y, depth_effort_.force.z);
+       if (max_depth_effort_ > 0.0 && depth_force_pool.norm() > max_depth_effort_) {
+           depth_force_pool = depth_force_pool * (max_depth_effort_ / depth_force_pool.norm());
```
WilliamZhang2205, line 179: how are these max_*_effort determined? Could they overlap on each other? What if the sum of the two clamped forces go over the total limit? From what I see the total forces arent clamped

ros2_ws/src/controls/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,140 @@
...
+        void TrajectoryPlanner::trajectory_timer_callback()
+        {
+                if (flip_in_progress_)
+                {
+                        rclcpp::Time now = this->now();
+                        double elapsed_sec = (now - flip_start_time_).seconds();
+
+                        if (elapsed_sec >= total_duration_)
+                        {
+                                // Flip complete
```
WilliamZhang2205, line 85: in the readme theres a note about flips of less than 15s not being supported. Does that encompass cases of 1 fast flip and like 15 flips in 15s?

ros2_ws/src/controls/src/trajectory_planner.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+                        q_v1v2.w() = std::cos(theta / 2.0);
+                        q_v1v2.vec() = flip_axis_ * std::sin(theta / 2.0);
+
+                        quatd q_target = q_start_ * q_v1v2;
+
+                        float s_dot = minimum_jerk_derivative(u);
+                        Vec3 w_ref_v = s_dot * flip_axis_ * total_angle_rad_/ total_duration_;
+
+                        publish_attitude_reference(q_target, w_ref_v);
```
WilliamZhang2205, line 104: just confirming that i understand this and the attitude controller, the w_ref is not a target angular velocity. It is just a term that adjusts the second P controller's influence.

ros2_ws/src/propulsion/docs/AUV_Thrust_Allocation_Procedure.pdf
WilliamZhang2205, line 1: i might be lacking the theoretical background but what exactly is the difference between a and a*, u and u*?

ros2_ws/src/propulsion/propulsion/thrust_allocator.py
```diff
@@ -0,0 +1,175 @@
...
+        # -- LP Solver setup
+        self.P_lp = sp.csc_matrix((9, 9))
+
+        self.q_lp = np.zeros(9)
+        self.q_lp[8] = -1.0  # Maximize alpha (minimize -alpha)
+
+        # Uniform scaling: all 6 wrench components are scaled by alpha.
+        tau_placeholder = -np.ones((6, 1))
+
+        self.A_lp = sp.bmat([
```
WilliamZhang2205, line 39: Could this be in the doc, like the fusing of each matrix into a super block matrix? I think it is helpful to see the equations in their paper form instead of the sp library's syntax.

WilliamZhang2205 (changes requested):
Great stuff overall! I still need to review teleop part, as always we will only merge after works in pool 👍

ros2_ws/src/controls/src/trajectory_planner.cpp
nadimasmar: My bad on the lack of documentation. This values represent the terms of the [minimum jerk polynomial](https://arxiv.org/pdf/2102.07459). This polynomial outputs the smoothest trajectory in normalized space ( 0 < x < 1) and time ( 0 < t < 1). 

I dont have much time to write detailed documentation at the moment. If it is successful and this PR is merged, I can issue a seperate PR to add missing documentation.

ros2_ws/src/controls/src/trajectory_planner.cpp
nadimasmar: Same comment as above

ros2_ws/src/auv_msgs/msg/FlipCommand.msg
nadimasmar: Yes, it is possible.

ros2_ws/src/auv_msgs/msg/AttitudeReference.msg
nadimasmar: Done

ros2_ws/src/controls/include/controls/trajectory_planner.hpp
nadimasmar: naming changed. I dont see this a number that we will be changing, I prefer to keep it simple and not have it configurable.

ros2_ws/src/controls/src/superimposer.cpp
nadimasmar: These are there to prevent a singular controller from producing a very high force that would drown out the contributions from other controllers. They are non-essential safeguards are currently set to 0.0. 

Clamping and stauration prevention is done in the propulsion package.

ros2_ws/src/controls/src/trajectory_planner.cpp
nadimasmar: The fl;ip duration is the amount of time allowed to complete a singular flip. 

It encompasses any flip that has a period of less than 15 seconds. You can have 2 flips take 30 seconds no problem. But you cannot ask for any number of flips which have a duration of less than 15 seconds.

ros2_ws/src/controls/src/trajectory_planner.cpp
nadimasmar: It is both. The second P controller goes to zero if and only if the the angular velocity is equal to w_ref. 

This means that it is a target angular velocity, which affects the output of the second P controller,

ros2_ws/src/propulsion/propulsion/thrust_allocator.py
nadimasmar: Yes of course. I'll add more documentation once we clear out our priorities.

avyan-k (changes requested):
LGTM code wise, considering the discussion above, the comments about documentation and services can be put on the back burner for now, under the hope and assumption that they will be resolved in a future PR. 

I will also add a commit after this review changing some parts of the READMEs, mostly some small nitpicky things that I've changed as I read through it that don't require mention in the review itself. Please correct the runtime error potential bug, that is the main blocking request change for now.

Otherwise, great work!

ros2_ws/src/propulsion/docs/AUV_Thrust_Allocation_Procedure.pdf
avyan-k: In mathematical programming (i.e. linear programming and OSQP used here), the `*` superscript just refers to the optimal values for the variables such that the objective function is optimized and all constraints are satisfied (i.e. your goal). 

In this case, $\alpha^\ast$ is the value that minimizes $f(x)=-\alpha$ (i.e. largest scaling we can achieve on the wrench) and $\mathbf{u}^\ast$ is the vector with the smallest L2 norm (i.e. the thruster allocation with the smallest power consumption).

PR #133, opened 2026-09-19
Add automatic submodule safe directory addition

WilliamZhang2205:
Instead of manually config adding each submodule as a safe directory, parse each submodule from the .gitmodules folder and add them as safe directory.

More dynamic.


kaidalisohaib (approved):
lgtm

prismaticbridge (approved):
lgtm

